PR #15, opened 2017-12-05
use requests before read_csv to get auth

rsignell-usgs:
This a workaround to access authenticated ERDDAP servers.
Since the pull requests for pandas to support authentication on `read_csv` requests (https://github.com/pandas-dev/pandas/pull/17087) has not yet been merged, we can read the URL using requests and then pass to pandas using StringIO.

### DIFF
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -273,8 +273,12 @@ class ERDDAP(object):
             ['latitude', 'longitude', 'time', 'depth']
 
         """
+        import requests
+        from io import StringIO
+            
         try:
             import pandas as pd
+
         except ImportError:
             raise ImportError('pandas is needed to use `get_var_by_attr`.')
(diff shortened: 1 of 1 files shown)


rsignell-usgs:
I tested this fix with this notebook:
https://github.com/reproducible-notebooks/ERDDAP_timeseries_explorer/blob/osu_auth/ERDDAP_timeseries_explorer.ipynb

erddapy/erddapy.py
```diff
@@ -1,5 +1,7 @@
 from __future__ import (absolute_import, division, print_function)
 
+import requests
```
ocefpaf: @rsignell-usgs if we leave this inside the function it would be to make it optional only when using `get_var_by_attr`.
In my next version `requests` will be mandatory , so instead of adding a try/except here I just moved it to the top already.

PR #17, opened 2017-12-08
Simplify parse_dates

rsignell-usgs:
Use timestamp() to simplify parse_dates function


ocefpaf:
I'll fix the coding standard lint in another PR. Thanks!

Issue #30, opened 2018-02-05
ERDDAPY returning variables that don't exist


rsignell-usgs:
@ocefpaf, I tried running ERDDAPY intro notebook for the OOI endpoint, and I think
there is a bug in `get_var_by_attr`.

In [this notebook](https://gist.github.com/anonymous/886a72417ee37ab94b9c4a8d479478e3) 
in cell [16] the names that come back include `density` and
`practical_salinity`, which do not exist in the dataset, which you can confirm here:

https://erddap-uncabled.oceanobservatories.org/uncabled/erddap/info/CP05MOAS-GL336-02-FLORTM000-flort_m_glider_instrument-telemetered-deployment0005-tabledap/index.html

ocefpaf:
Fixed in #31

PR #31, opened 2018-02-05
remove loop in `get_var_by_attr`

rsignell-usgs:
This fixes issue #30.


ocefpaf:
The two hardest problems in computer science: caching invalidation and
naming things  ;-p

PR #33, opened 2018-03-08
Fix IFREMER URL and add SalishSeaCast to servers.

circularpenguin:
HI! I would like to add SalishSeaCast to the list of servers and I also corrected the IREMER's URL (it's shown to be incorrect [here](https://github.com/ropensci/rerddap/pull/69)). Let me know if there is anything else I should do!


ocefpaf:
Thanks @circularpenguin!

PR #56, opened 2018-10-05
Test Python 3.7

ocefpaf:
Not sure if we have all the dependencies yet...


ocefpaf:
I guess we have all the dependencies.

PR #70, opened 2019-01-31
Increase test coverage...

ocefpaf:
... to avoid breakages when adding new features.


ocefpaf:
I cannot reproduce the `coding_standards` failures locally and they are braking before the test in the CIs :-/

PR #94, opened 2019-09-10
Pass protocol to the URL builder...

ocefpaf:
..  and remove wrong docs.

Ping @mwengren.

### DIFF
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -7,6 +7,7 @@ Pythonic way to access ERDDAP data
 
 import copy
 import functools
+
 from urllib.parse import quote_plus
 
 import pandas as pd
@@ -105,6 +106,7 @@ class ERDDAP(object):
         self,
         response=None,
         search_for=None,
+        protocol=None,
         items_per_page=1000,
         page=1,
         **kwargs,
@@ -122,8 +124,6 @@ class ERDDAP(object):
                 - To exclude datasets with a specific word, use `-excludedWord`.
                 - To exclude datasets with a specific phrase, use `-"excluded phrase"`
                 - Searches are not case-sensitive.
-                - To find just grid or just table datasets, include `protocol=griddap`
-                    or `protocol=tabledap` in your search.
                 - You can search for any part of a word. For example,
                     searching for `spee` will find datasets with `speed` and datasets with
                     `WindSpeed`
@@ -175,8 +175,12 @@ class ERDDAP(object):
         if max_time:
             kwargs.update({"max_time": parse_dates(max_time)})
 
-        default = "(ANY)"
+        protocol = protocol if protocol else self.protocol
         response = response if response else self.response
+        if protocol:
+            kwargs.update({"protocol": protocol})
+
+        default = "(ANY)"
         url = base.format(
             server=self.server,
             response=response,
--- a/tests/test_url_builder.py
+++ b/tests/test_url_builder.py
@@ -58,7 +58,10 @@ def test_search_url_valid_request(e):
     assert options.pop("maxTime") == str(parse_dates(max_time))
     assert options.pop("itemsPerPage") == str(1000)
     for k, v in options.items():
-        assert v == "(ANY)"
+        if k == "protocol":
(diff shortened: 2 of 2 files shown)


erddapy/erddapy.py
```diff
@@ -122,8 +124,6 @@ def get_search_url(
                 - To exclude datasets with a specific word, use `-excludedWord`.
                 - To exclude datasets with a specific phrase, use `-"excluded phrase"`
                 - Searches are not case-sensitive.
-                - To find just grid or just table datasets, include `protocol=griddap`
-                    or `protocol=tabledap` in your search.
```
ocefpaf: This is from ERDDAP docs but it does not seems to work for me.

mwengren: @ocefpaf Maybe check with @BobSimons to make sure the functionality is as expected.  Maybe it only works on certain ERDDAP versions.

When we were working on the OHW query tool a few weeks ago, we had to do a secondary filter to remove results without tabledap endpoints when we thought we had the filter configured.  I assumed it was just because erddapy was submitting 'protocol=(ANY)' rather than 'protocol=tabledap'.  Without testing every server, I did get it to work correctly on this one:

Returns two datasets w/out tabledap endpoints:
https://apdrc.soest.hawaii.edu/erddap/search/advanced.csv?page=1&itemsPerPage=1000&protocol=tabledap&cdm_data_type=(ANY)&institution=(ANY)&ioos_category=(ANY)&keywords=(ANY)&long_name=(ANY)&standard_name=sea_water_practical_salinity&variableName=(ANY)&minLon=-75.857066&maxLon=-71.28504800000002&minLat=33.19150500000001&maxLat=39.689154&minTime=1563926400.0&maxTime=1577750400.0

Returns 500 status code for no matching datasets: 
https://apdrc.soest.hawaii.edu/erddap/search/advanced.csv?page=1&itemsPerPage=1000&protocol=tabledap&cdm_data_type=(ANY)&institution=(ANY)&ioos_category=(ANY)&keywords=(ANY)&long_name=(ANY)&standard_name=sea_water_practical_salinity&variableName=(ANY)&minLon=-75.857066&maxLon=-71.28504800000002&minLat=33.19150500000001&maxLat=39.689154&minTime=1563926400.0&maxTime=1577750400.0

ocefpaf: > @ocefpaf Maybe check with @BobSimons to make sure the functionality is as expected. Maybe it only works on certain ERDDAP versions.
> 
> When we were working on the OHW query tool a few weeks ago, we had to do a secondary filter to remove results without tabledap endpoints when we thought we had the filter configured. I assumed it was just because erddapy was submitting 'protocol=(ANY)' rather than 'protocol=tabledap'. Without testing every server, I did get it to work correctly on this one:


Specifying `protocol=tabledap` on the search query works with some servers. However, adding `&protocol=table` directly (and this was buggy in erddapy before) also works. I prefer the latter b/c we can use the instance `protocol` property instead of asking the user to re-pass the argument. Also, users can override that with a keyword.

Are you OK with this PR?

erddapy/erddapy.py
```diff
@@ -175,8 +175,12 @@ def get_search_url(
         if max_time:
             kwargs.update({"max_time": parse_dates(max_time)})
 
-        default = "(ANY)"
+        protocol = protocol if protocol else self.protocol
         response = response if response else self.response
+        if protocol:
+            kwargs.update({"protocol": protocol})
```
ocefpaf: @mwengren the bug was here. The `protocol` variable was not passed to the URL builder.

Another way to handle this would be to let the user pass this in the search query but, while that works great for the HTML version of the search, using the `protocol` property set in the class makes more sense.

erddapy/erddapy.py
ocefpaf: It did work with another server. As Micah mentioned below it may be version specific. Still, we don't need that in `erddapy`. For API consistence let's get that information from the main class.

mwengren:
Looks good to me!  Sorry missed your question yesterday.

PR #97, opened 2019-10-10
Update README.md

Bobfrat:
data.ioos.us/gliders  will redirect appropriately but better to advertise this URL.


ocefpaf:
Thanks @Bobfrat!

PR #99, opened 2019-10-29
add server-side function distinct

### LINKED ISSUE #98: Support ERDDAP "distinct()" server-side function
erddapy doesn't currently support ERDDAP "Server-side functions". The only such functions I'm aware of are `distinct()` and various ordering functions. The latter don't really seem necessary in the context of erddapy and pandas, b/c sorting can be done client-side via Pandas. But `distinct()` does make a difference, as it can make the response much smaller, depending on the data.

It's actually very easy to use distinct with erddapy as-is. Once I have the request ready to go, I pass this url to pandas read_csv:
```python
e.get_download_url() + '&distinct()'
```

But figuring out how to do this takes a fair bit of work. It'd save users some effort if this was built into erddapy. I could be persuaded to try to add it myself and submit a PR, but not this month ...

ocefpaf:
Closes #98 

Work in progress, needs:

- [ ] docstrings
- [ ] notebook example
- [ ] expand the functionality too all returns

@emiliom this implements `distinct` only for `download_url` but it seems that it makes sense to add this too all URLs. Not sure what is the most sane way to do that yet.

PS: do you have a nice example of this in action for the notebook/documentation?

### DIFF
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -35,7 +35,7 @@ def _distinct(url: str, **kwargs) -> str:
     See https://coastwatch.pfeg.noaa.gov/erddap/tabledap/documentation.html#distinct
 
     """
-    distinct = kwargs.get("distinct")
+    distinct = kwargs.pop("distinct")
     if distinct:
         return f"{url}&distinct()"
     return url
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -37,7 +37,7 @@ def _distinct(url: str, **kwargs) -> str:
 
     """
     distinct = kwargs.pop("distinct", False)
-    if distinct:
+    if distinct is True:
         return f"{url}&distinct()"
     return url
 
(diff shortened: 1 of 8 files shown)


erddapy/erddapy.py
```diff
@@ -22,6 +22,14 @@
 )
 
 
+def _distinct(url: str, **kwargs) -> str:
+    """Apply a server side function."""
+    distinct = kwargs.get("distinct")
+    if distinct:
```
emiliom: For what it's worth, this is testing for `distinct` having any value that produces a True in the if test, not just the value `True`. So, `distinct='abc'` will also result in the test being True.

emiliom:
Looks good, other than my minor comment.

I haven't explored other urls other than `download_url`.

erddapy/erddapy.py
ocefpaf: Yep. Still factoring it out in its own module so we can use typing and make it strict Boolean.

emiliom:
Looking good!

PR #101, opened 2019-12-03
pass requests_kwargs to _check_url_response [fixes #100]

### LINKED ISSUE #100: _check_url_response() should pass requests_kwargs
While trying to connect to a ERDDAP server that had basic authentication set up I found that _check_url_response() was not passing any previously supplied requests_kwargs which are needed to pass long the authentication information.

This is already being done for calls to urlopen()

jmunroe:
Closes #100

### DIFF
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -203,7 +203,7 @@ class ERDDAP(object):
             searchFor=search_for,
         )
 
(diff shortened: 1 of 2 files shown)


ocefpaf:
Thanks @jmunroe! I'll probably re-factor this once pandas can use requests as an optional dependency.

PR #104, opened 2020-02-12
Drop versioneer, use setuptools_scm and avoid setup.py as much as possible

ocefpaf:
In this PR

- moved away from `versioneer`, which is deprecated, to `setuptools_scm`
- moved most of the config stuff to `setup.py`


ocefpaf:
OK. Let's give this a try!

PR #138, opened 2020-09-18
fixes 137

ocefpaf:
@abkfenris this was broken since I added the in-memory netCDF. However, the real mystery here is why the in-memory stopped working on that server. I'm still investigating that b/c I want to add tests for both cases.


abkfenris:
I suggest [VCRpy](https://vcrpy.readthedocs.io/en/latest/) for catching fixtures of issues like this, so you don't have to depend on the server to keep making the same mistake on demand.

I usually set it up so it doesn't have to match quite so strictly

https://github.com/gulfofmaine/buoy_barn/blob/56aec9ce3d7a40b6ad85b4b9a7f53bb813a7f897/app/deployments/tests/vcr.py#L1-L4

Then decorate the test.

https://github.com/gulfofmaine/buoy_barn/blob/master/app/deployments/tests/test_tasks.py#L116-L124

[pytest-vcr](https://pytest-vcr.readthedocs.io/en/latest/) also can give you a little more pytestish interface.

I had been thinking about going through and adding VCR tests at some point. pytest-vcr might be nice because you can also do a test run with test disabled quickly.

ocefpaf:
Yep. +1 to vcrpy!

ocefpaf:
BTW, in-memory is working again for that server but the fix here is still necessary for the cases when it will fail. I'll leave the addition of better test with vcr to another PR.

ocefpaf:
> I'll leave the addition of better test with vcr to another PR.

Actually it was easier than I thought with pytest-vcr.

PR #145, opened 2020-10-05
Refactor

ocefpaf:
We should avoid having a `utilities.py` module in our projects b/c it leads to a bloated module without a clear purpose. This refactor splits it into more meaningful modules that will hopefully improve maintenance.


tests/test_erddapy.py
```diff
@@ -14,10 +15,7 @@ def test_erddap_requests_kwargs():
     timeout_seconds = 1  # request timeout in seconds
     slowwly_milliseconds = (timeout_seconds + 1) * 1000
     slowwly_url = (
-        "http://slowwly.robertomurray.co.uk/delay/"
```
ocefpaf, line 17: @abkfenris I had to change this b/c the service was down. Not sure if temporarily or permanently. Maybe we can test this without relying on this kind of service. (Ironically this library is mostly a "server" test ;-p)

abkfenris: Maybe it's just being extra slow....

It looks like there was some experimentation with adding a [timeout in responses](https://github.com/getsentry/responses/pull/115) but that stalled out.

Most mocks are suggesting throwing a `ConnectionTimeout` directly, rather than seeing if the requests actually times out, it's forcing the matter sooner.

Doesn't VCR return immediately rather than simulating a delay?

ocefpaf: Not sure about VCR's details here. The cassette yaml file had the `http://slowwly.robertomurray.co.uk/delay/` in but it was still trying to access it and failing.

PR #147, opened 2020-10-06
cache urlopen

ocefpaf:
This should speed up returns when the URL and parameters are used.


ocefpaf:
I need to move the doc build to GH Actions to avoid Travis-CI time-outs :-/

Issue #161, opened 2020-11-19
Allow server-side functions within time parameter


thogar-computer:
Afternoon, 
currently if a user wants to call an URL like the following from erddapy;

http://erddap.sensors.ioos.us/erddap/tabledap/uk_gov_metoffice_62103.htmlTable?time,latitude,longitude,wind_speed,wind_from_direction&time=max(time)+0minutes

It will fail on [line 50 - erddapy.py](https://github.com/ioos/erddapy/blob/master/erddapy/erddapy.py#L50)

``` python
parse_date_time = parse_time_string(date_time)[0]
```

Even though this can be done from within pandas, these types of server-side functions greatly reduce the amount of data transmitted.  

As a starter, having a check on the value to see if it starts with max/min? I am not totally sure on the full list of functions...

If agreed, i am happy to pick this up :)

ocefpaf:
That would be a nice addition. If you want to send a PR it will be most welcomed!

thogar-computer:
cheers for the merge :) i will close this now

PR #162, opened 2020-11-23
Feature/server functions

thogar-computer:
Changes to allow the use of server side functions within the URL parameters as built up from within erddapy

as mentioned in #161

This will work across any URL parameter time or otherwise. After looking over the code it was felt best to add in a new attribute within the erddap class modelled on constraints, thus skipping the need for complex string/time checks and if statements.

### DIFF
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -134,6 +134,7 @@ class ERDDAP:
 
         # Initialized only via properties.
         self.constraints: Optional[Dict] = None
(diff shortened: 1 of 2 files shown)


ocefpaf (approved):

ocefpaf:
@thogar-computer PR #163 fixed the pre-commit issues. Do you mind rebasing this so we can run those tests here? If that is too much don't worry, we can merge without it.

PR #182, opened 2021-06-14
formulate griddap download url

callumrollo:
Implementation is a bit messy. Requires the user to call two functions of ERDDAP class. To test:
```python
e = ERDDAP(
    server="CSWC",
    protocol="griddap",
    response='csvp'
)
e.dataset_id = "usgsCeCrm3"

e.griddap_initialise()

# make changes to e.constraints and e.variables as desired

print(e.get_download_url())
```

### DIFF
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -451,8 +451,9 @@ class ERDDAP:
         if not protocol:
             raise ValueError(f"Please specify a valid `protocol`, got {protocol}")
 
-        if protocol == "griddap":
+        if protocol == "griddap" and constraints is not None and variables is not None:
             # Check that dimensions, constraints and variables are valid for this dataset
+
             _griddap_check_constraints(self.constraints, self._constraints_original)
             _griddap_check_variables(self.variables, self._variables_original)
             download_url = [
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -5,6 +5,7 @@ Pythonic way to access ERDDAP data
 
 import copy
 import functools
+import urllib
 from datetime import datetime
 from typing import Dict, List, Optional, Tuple, Union
 from urllib.parse import quote_plus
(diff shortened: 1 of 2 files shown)


erddapy/erddapy.py
```diff
@@ -65,6 +66,50 @@ def parse_dates(date_time: Union[datetime, str]) -> float:
...
 
 
+def _griddap_get_constraints(dataset_url, step=1000):
+    """
+    Fetch metadata of griddap dataset and set initial constraints
+    Step size is applied to all dimensions
+
+    """
+    dds_url = dataset_url + ".dds"
+    with urllib.request.urlopen(dds_url) as url:
```
ocefpaf, line 76: We already depend on `requests` so using that instead of the standard lib is fine. Also requests handles some exotic ERDDAP URLs that urllib does not.

callumrollo:
Is this implementation ok? If so, I'll add some tests and try to clean up the UI a bit. 

Ideally the metadata fetching with the .dds file would happen as soon as the user supplies a griddap dataset id. If we can detect this, we could save the call to `griddap_initialise`

Also, is it good to have a default dictionary of subsetting that works out the box, or should we get the user to build one from scratch? Currently it defaults to the full range of all dimensions, with a step of 1000 to keep the file size low

erddapy/erddapy.py
```diff
@@ -65,6 +65,51 @@ def parse_dates(date_time: Union[datetime, str]) -> float:
...
 
 
+def _griddap_get_constraints(dataset_url, step=1000):
+    """
+    Fetch metadata of griddap dataset and set initial constraints
+    Step size is applied to all dimensions
+
+    """
+
+    dds_url = dataset_url + ".dds"
```
ocefpaf, line 75: Let's use f-strings, they are faster and in most cases more readable. It would be something like this:

```python
f"{dataset_url}.dds"
```

erddapy/erddapy.py
```diff
@@ -147,12 +192,22 @@ def __init__(
...
         self.variables: Optional[ListLike] = None
+        self.dim_names: Optional[ListLike] = None
 
         # Caching the last `dataset_id` and `variables` list request for quicker multiple accesses,
         # will be overridden when requesting a new `dataset_id`.
         self._dataset_id: OptionalStr = None
         self._variables: Dict = {}
 
+    def griddap_initialise(self):
+        if self.protocol == "griddap":
```
ocefpaf, line 203: Should probably throw an error is the protocol is not griddap. What do you think?

ocefpaf:
> Is this implementation ok? If so, I'll add some tests and try to clean up the UI a bit.

Looks like you are in the right path. To be honest I don't have a clear vision for a nice API but we can merge this one and keep iterating until we are happy with it.

> Ideally the metadata fetching with the .dds file would happen as soon as the user supplies a griddap dataset id. If we can detect this, we could save the call to `griddap_initialise`

That would be nice. We can make it lazy enough to trigger that only when a final URL is requested, right?

> Also, is it good to have a default dictionary of subsetting that works out the box, or should we get the user to build one from scratch? Currently it defaults to the full range of all dimensions, with a step of 1000 to keep the file size low

What is ERDDAP's default? If they don't have one this is fine as long as the large step will always produce a valid slice. Like 0:1000:10 should yield at least the first data point.

callumrollo:
> > Ideally the metadata fetching with the .dds file would happen as soon as the user supplies a griddap dataset id. If we can detect this, we could save the call to `griddap_initialise`
> 
> That would be nice. We can make it lazy enough to trigger that only when a final URL is requested, right?
>
If we only trigger it when the final URL is requested, the user won't be able to make any changes to the request (changing limits, dropping unwanted variables etc.) I think it needs to be triggered as soon as the protocol='griddap' and a valid dataset id are supplied. Unsure how to implement though
> > Also, is it good to have a default dictionary of subsetting that works out the box, or should we get the user to build one from scratch? Currently it defaults to the full range of all dimensions, with a step of 1000 to keep the file size low
> 
> What is ERDDAP's default? If they don't have one this is fine as long as the large step will always produce a valid slice. Like 0:1000:10 should yield at least the first data point.

afaik there is no default. This method will work for all datasets though, if they have fewer than 1000 points in a dimension, it will return only the first point. We can always change the default stride later. We'll need to add some tests to make sure the user hasn't supplied invalid variables or min/max dimenstion values out of range

callumrollo:
I'm getting errors from mypy in the pre-commit:

<pre>mypy.....................................................................<span style="background-color:#CC0000">Failed</span>
<font color="#9F9F9D">- hook id: mypy</font>
<font color="#9F9F9D">- exit code: 1</font>

erddapy/erddapy.py:442: <font color="#EF2929"><b>error:</b></font> Item <b>&quot;None&quot;</b> of <b>&quot;Union[List[str], Tuple[str], None]&quot;</b> has no attribute <b>&quot;__iter__&quot;</b> (not iterable)
erddapy/erddapy.py:444: <font color="#EF2929"><b>error:</b></font> Item <b>&quot;None&quot;</b> of <b>&quot;Union[List[str], Tuple[str], None]&quot;</b> has no attribute <b>&quot;__iter__&quot;</b> (not iterable)
erddapy/erddapy.py:446: <font color="#EF2929"><b>error:</b></font> Value of type <b>&quot;Optional[Dict[Any, Any]]&quot;</b> is not indexable
<font color="#EF2929"><b>Found 3 errors in 1 file (checked 1 source file)</b></font>
</pre>

I haven't worked with this type checking stuff before. How should I go about fixing it?

callumrollo:
There are more checks/input sanitising to be done, but I'm happy to merge this one now and refine later

callumrollo:
Other than the known issue with mypy, this should be good to merge. Test failures were caused by the parser trying to build a griddap query with no constraints or variables.

PR #185, opened 2021-06-22
Improve docstring for `get_categorize_url()` function

vinisalazar:
Hi,

As an ERDDAP beginner, I was having some trouble discovering what are valid attributes to categorize by. After realising that I could find them at the `erddap/categorize/index.html` page, I decided to improve the `get_categorize_url()` docstring for other users that may run into the same problem.

Summary of changes:
  - In `get_categorize_url()` function: Add a line to the docstring explaining that valid attributes can be found in the categorize page.

Best,
Vini

### DIFF
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -306,7 +306,7 @@ class ERDDAP:
 
         Args:
             categorize_by: a valid attribute, e.g.: ioos_category or standard_name.
-                Valid attributes are shown in the '<YOUR-ERDDAP-URL>/erddap/categorize' page.
+                Valid attributes are shown in the https://coastwatch.pfeg.noaa.gov/erddap/categorize page.
             value: an attribute value.
             response: default is HTML.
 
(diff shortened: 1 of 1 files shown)


erddapy/erddapy.py
```diff
@@ -306,6 +306,7 @@ def get_categorize_url(
 
         Args:
             categorize_by: a valid attribute, e.g.: ioos_category or standard_name.
+                Valid attributes are shown in the '<YOUR-ERDDAP-URL>/erddap/categorize' page.
```
ocefpaf, line 309: The https://coastwatch.pfeg.noaa.gov/erddap/categorize server is the "gold standard" ERDDAP server, so maybe we can use that URL instead.

vinisalazar: Thank you @ocefpaf

ocefpaf:
Hope it was the first of many ;-p

Thanks @vinisalazar!

PR #186, opened 2021-06-23
fix type output of _griddap_get_constraints

ocefpaf:
@callumrollo I fixed the output type info but there are still a few other places to fix:

```
erddapy/erddapy.py:460: error: Argument 1 to "_griddap_check_constraints" has incompatible type "Optional[Dict[Any, Any]]"; expected "Dict[Any, Any]"
erddapy/erddapy.py:461: error: Argument 1 to "_griddap_check_variables" has incompatible type "Union[List[str], Tuple[str], None]"; expected "List[Any]"
erddapy/erddapy.py:472: error: Item "None" of "Union[List[str], Tuple[str], None]" has no attribute "__iter__" (not iterable)
erddapy/erddapy.py:474: error: Item "None" of "Union[List[str], Tuple[str], None]" has no attribute "__iter__" (not iterable)
erddapy/erddapy.py:476: error: Value of type "Optional[Dict[Any, Any]]" is not indexable
```

I'll send a few extra commits here to fix those ASAP but feel free to tackle them if you want to.


ocefpaf:
We are down to one!

```
erddapy/erddapy.py:474: error: Item "None" of "Union[List[str], Tuple[str], None]" has no attribute "__iter__" (not iterable)
Found 1 error in 1 file (checked 13 source files)
```

erddapy/erddapy.py
```diff
@@ -65,7 +65,10 @@ def parse_dates(date_time: Union[datetime, str]) -> float:
     return parse_date_time.timestamp()
 
 
-def _griddap_get_constraints(dataset_url: str, step: int = 1000) -> [Dict, List, List]:
+def _griddap_get_constraints(
+    dataset_url: str,
+    step: int = 1000,
+) -> Tuple[Dict, List, List]:
```
ocefpaf, line 71: @callumrollo the return is a tuple with those types. That was one of the type failures we had and I missed it when reviewing your PR. Sorry.

erddapy/erddapy.py
```diff
@@ -118,7 +121,7 @@ def _griddap_check_constraints(user_constraints: Dict, original_constraints: Dic
         )
 
 
-def _griddap_check_variables(user_variables: List, original_variables: List):
+def _griddap_check_variables(user_variables: ListLike, original_variables: ListLike):
```
ocefpaf, line 124: This will take List or Tuple now.

erddapy/erddapy.py
```diff
@@ -454,8 +457,8 @@ def get_download_url(
         if protocol == "griddap" and constraints is not None and variables is not None:
             # Check that dimensions, constraints and variables are valid for this dataset
 
-            _griddap_check_constraints(self.constraints, self._constraints_original)
-            _griddap_check_variables(self.variables, self._variables_original)
```
ocefpaf, line 458: We parse these above and, inside this function, I believe we do not want to use the the one from the class to avoid ignoring the user input.

ocefpaf:
@callumrollo I narrowed it down to this part:

```python
        (
            self.constraints,
            self.dim_names,
            self.variables,
        ) = _griddap_get_constraints(metadata_url)
```

again, my bad for not paying closer attention to the CI errors.

There we are clobbering the `self.constraints`, `self.dim_names`, and `self.variables`. The type check fails b/c None is possible and they are optional in the class. So we need to get non "self.` version of those first and update the class attribute only when it makes sense.

Do you want to tackle that? If so I'll merge this PR and let you work on it.

callumrollo:
I'll have a go! Still finding Type checking a little mystifying, but I'll learn from the examples you've made. Good to merge this one now

PR #187, opened 2021-06-27
Case-insensitive server names

callumrollo:
Makes use of server short names case insensitive, so that POLARWATCH, polarwatch and PolarWatch will all point to https://polarwatch.noaa.gov/erddap

As suggested over on gliderpy https://github.com/ioos/gliderpy/issues/38#issuecomment-737178290


callumrollo:
@ocefpaf 
Refactored to lower keys at server dict creation. Ready to merge

PR #189, opened 2021-06-27
check for dataset_id in griddap_initialize

callumrollo:
raises a ValueError if user has not set dataset_id


callumrollo:
@ocefpaf  Added line to accept dataset_id or self.dataset_id
Ready for merge

PR #192, opened 2021-06-30
default stride to 1 in griddap

callumrollo:
Recreate default griddap behaviour when subsetting gridded data

PR #194, opened 2021-07-02
remove unused deps

ocefpaf:
@callumrollo also using this one to test the CI failure you are experimenting in #192.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,9 +17,10 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Setup Conda
-      uses: s-weigand/setup-conda@v1
+      uses: s-weigand/setup-conda@v1.0.5
       with:
         activate-conda: false
+        update-conda: true
         conda-channels: conda-forge
 
     - name: Python ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Setup Conda
-      uses: s-weigand/setup-conda@v1
+      uses: s-weigand/setup-conda@v1.0.5
```
ocefpaf, line 20: @callumrollo looks like it is the latest release of this GHA (1.0,6). Pinning to the previous one fixes it. I'll push that change to your branch.

callumrollo:
@ocefpaf can we merge this? Then I can update the two other PRs

.github/workflows/tests.yml
```diff
@@ -17,22 +17,28 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Setup Conda
-      uses: s-weigand/setup-conda@v1
+      #uses: s-weigand/setup-conda@v1.0.5
+      uses: s-weigand/setup-conda@fix-env-activation
```
s-weigand, line 21:
```suggestion
      uses: s-weigand/setup-conda@fix-env-activation-dist
```
If this fixes the issue with the activation of the `TEST` env.
I will release a new version of `setup-conda` which should fix the usage of the `v1` tag.
Thanks for testing ❤️ 
Ref:
https://github.com/s-weigand/setup-conda/issues/125

ocefpaf: Sorry for the missing `-dist` there. I tried that using the mobile web interface during the weekend, it was doomed to fail ;-p

Thanks for the quick fix. Let's see how the CIs go.

ocefpaf: @s-weigand  works like a charm! Thanks!! I'll wait for the release to merge this one.

ocefpaf:
> @ocefpaf can we merge this? Then I can update the two other PRs

Not right now. Let's wait for an official release of the GHA. Meanwhile I'm OK downgrading to the previous version.

s-weigand:
Give me 5min and you will have the release 😄 
Thanks again for testing!

ocefpaf:
Thanks so much for the quick fix @s-weigand!

PR #193, opened 2021-06-30
started griddap notebook

callumrollo:
This will eventually close #191 need to figure out how the notebook becomes part of the documentation first though

PR #195, opened 2021-07-05
try latest mypy

ocefpaf:
Latest mypy does not install all stub by default and installing them in the conda env is not "visible" in for the pre-commit interpreter. Let's see if we can make it work with this setup.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -82,7 +82,6 @@ repos:
   hooks:
     - id: prenotebook
 
-
 ci:
     autofix_commit_msg: |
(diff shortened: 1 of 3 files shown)


ocefpaf:
@callumrollo I reported the prenotebook failure upstream. See https://github.com/deathbeds/prenotebook/issues/6

callumrollo:
Cool. Though it's weird it only fails for some PRs. e.g. #193 is passing atm

ocefpaf:
> Cool. Though it's weird it only fails for some PRs. e.g. #193 is passing atm

I think you nailed it. That one is passing b/c it has a notebook. I guess it is not smart enough to not run when it shouldn't.

Issue #32, opened 2018-02-06
Griddap params


flackdl:
Hey there, 

I'm hoping to use your nifty package but am running into an issue.  You should note I don't have a climate science background.  

When I run the following I receive an error, which, it seems like the generated URL doesn't conform to the [griddap documentation](https://data.ioos.us/gliders/erddap/griddap/documentation.html) but instead to [tabledap](https://data.ioos.us/gliders/erddap/tabledap/documentation.html).

```python
from erddapy import ERDDAP

e = ERDDAP(
    server='http://oos.soest.hawaii.edu/erddap',
    protocol='griddap',
    response='csv',
    dataset_id='NCEP_Global_Best',
    constraints={
        'time>=': '2018-02-08',
    },
    variables=['time', 'longitude', 'latitude'],
)

# http://oos.soest.hawaii.edu/erddap/griddap/NCEP_Global_Best.csv?time,longitude,latitude&time%3E=1518048000.0
print(e.get_download_url())
```
Outputs:
```
Traceback (most recent call last):
  File "/tmp/a.py", line 14, in <module>
    print(e.get_download_url())
  File "/home/danny/.virtualenvs/cwwed-env/lib/python3.6/site-packages/erddapy/erddapy.py", line 143, in get_download_url
    constraints=constraints,
  File "/home/danny/.virtualenvs/cwwed-env/lib/python3.6/site-packages/erddapy/url_builder.py", line 183, in download_url
    return _check_url_response(url)
  File "/home/danny/.virtualenvs/cwwed-env/lib/python3.6/site-packages/erddapy/utilities.py", line 101, in _check_url_response
    r.raise_for_status()
  File "/home/danny/.virtualenvs/cwwed-env/lib/python3.6/site-packages/requests/models.py", line 935, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http://oos.soest.hawaii.edu/erddap/griddap/NCEP_Global_Best.csv?time,longitude,latitude&time%3E=1518048000.0

```
If I use the [Data Access Form](http://oos.soest.hawaii.edu/erddap/griddap/NCEP_Global_Best.html) and only add the *time* constraint, it generates a url like:
 
http://oos.soest.hawaii.edu/erddap/griddap/NCEP_Global_Best.csv?time[(2018-02-01):1:(2018-02-13T12:00:00Z)] which works.

[...]

Issue #190, opened 2021-06-28
Multiple ERDDAP search


ocefpaf:
We could do an async search on a pre-determined list or maybe wrap an interface for [1].

@callumrollo is that something you would like to tackle?

[1] https://coastwatch.pfeg.noaa.gov/erddap/download/SearchMultipleERDDAPs.html

ocefpaf:
Here is a hacky example on how to do that: https://nbviewer.jupyter.org/gist/ocefpaf/74f91d9aa9c4fa726bfba171b3edcadf

callumrollo:
I tried integrating the extra checks into `erddapy.url_handling._urlopen`, but multiprocessing did not like the lru cache and kwargs, caused problems with the pickling. I think the least bad aproach would be a new function multi_urlopen for this specific purpose. I'll create a minimum working example

PR #199, opened 2021-07-17
multi erddap search added

callumrollo:
First attempt at resolving  #190


callumrollo:
`ERDDAP.search_all_servers` searches multiple servers for a user-supplied string. It returns a dataframe the dataset name, ID, institution and server url. All a user needs to make data queries against those datasets. The user can supply a list of server urls, or it tries all the servers in `erddapy.servers`

erddapy/erddapy.py
```diff
@@ -136,6 +138,22 @@ def _griddap_check_variables(user_variables: ListLike, original_variables: ListL
         )
 
 
+def parse_results(url: str, key: str, protocol="tabledap") -> Optional[Dict]:
+    """
+    Parse search results from multiple servers
+    """
+    data = multi_urlopen(url)
```
abkfenris, line 145: Is this parsing results, or is it fetching?

erddapy/erddapy.py
```diff
@@ -371,6 +389,34 @@ def get_search_url(
...
+                server: f'{server}search/index.csv?page=1&itemsPerPage=100000&searchFor="{query}"'
+                for server in servers_list
+            }
+        else:
+            urls = {
+                key: f'{server.url}search/index.csv?page=1&itemsPerPage=100000&searchFor="{query}"'
+                for key, server in servers.items()
+            }
+        num_cores = multiprocessing.cpu_count()
+        returns = Parallel(n_jobs=num_cores)(
```
abkfenris, line 411: It might make sense to make the parallelized searching optional. I haven't explored it in this case, but there are enough different ways to do multiprocessing in Python, that using joblib may limit how erddapy can be used in some environments (say with asyncio). 

One way to do this would be to decompose the function a into URL generation, parsing a single result, and combining all the results into a dataframe functions. Then they can be combined into serial and parallel search functions (or one function with a kwarg to switch between the two). Those functions then provide a blueprint for someone to reuse the decomposed functions in other multiprocessing methods.

erddapy/erddapy.py
```diff
@@ -371,6 +389,34 @@ def get_search_url(
...
+            delayed(parse_results)(url, key, protocol="tabledap")
+            for key, url in urls.items()
+        )
+        dfs = [x for x in returns if x is not None]
+        df_all = pd.concat([list(df.values())[0] for df in dfs])
+        df_all.reset_index(drop=True, inplace=True)
+        return df_all
+
```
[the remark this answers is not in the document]
callumrollo: Yes, that would make a lot more sense. I've moved it out in commit 4ab1e5aec767178abf2d02ec3157e35b44129fd8

erddapy/erddapy.py
```diff
@@ -371,6 +389,34 @@ def get_search_url(
...
+                for server in servers_list
+            }
+        else:
+            urls = {
+                key: f'{server.url}search/index.csv?page=1&itemsPerPage=100000&searchFor="{query}"'
+                for key, server in servers.items()
+            }
+        num_cores = multiprocessing.cpu_count()
+        returns = Parallel(n_jobs=num_cores)(
+            delayed(parse_results)(url, key, protocol="tabledap")
```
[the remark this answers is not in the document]
callumrollo: That's a good point. Protocol added as kwarg in 4ab1e5a

erddapy/erddapy.py
callumrollo: This has been renamed to fetch_results in 4ab1e5a

erddapy/erddapy.py
callumrollo: Yes, I've been looking into better ways to parallelise this. I thought I'd get a working version up for comment first, then try to optimise later. A big consideration will be what dependencies @ocefpaf is happy with for erddapy. 

Decomposing the functions so that others can perform their own parallelising is a good idea though. I'll refactor to accommodate this

callumrollo: I have made the parallelising via joblib a kwarg switch in `search_all_servers` in 949e4e0

callumrollo:
@abkfenris thank you for the review. I hope the subsequent commits have improved this PR. @ocefpaf could you look over this? I'm not sure why the pre-commit is failing again, it worked fine locally :/

erddapy/erddapy.py
```diff
@@ -136,6 +137,67 @@ def _griddap_check_variables(user_variables: ListLike, original_variables: ListL
...
+    if servers_list:
+        urls = {
+            server: f'{server}search/index.csv?page=1&itemsPerPage=100000&searchFor="{query}"'
+            for server in servers_list
+        }
+    else:
+        urls = {
+            key: f'{server.url}search/index.csv?page=1&itemsPerPage=100000&searchFor="{query}"'
+            for key, server in servers.items()
+        }
```
abkfenris, line 184: I'd suggest extracting the url generation into a separate function.

erddapy/erddapy.py
```diff
@@ -136,6 +137,67 @@ def _griddap_check_variables(user_variables: ListLike, original_variables: ListL
...
+    data = multi_urlopen(url)
+    if data is None:
+        return None
+    df = pd.read_csv(data)
+    try:
+        df.dropna(subset=[protocol], inplace=True)
+    except KeyError:
+        return None
+    df["Server url"] = url.split("search")[0]
+    return {key: df[["Title", "Institution", "Dataset ID", "Server url"]]}
```
abkfenris, line 153: I'd suggest splitting out the parsing of data from your data fetching function.

erddapy/erddapy.py
```diff
@@ -136,6 +137,67 @@ def _griddap_check_variables(user_variables: ListLike, original_variables: ListL
...
+        )
+        dfs = [x for x in returns if x is not None]
+    else:
+        dfs = []
+        for key, url in urls.items():
+            dfs.append(fetch_results(url, key, protocol=protocol))
+    df_all = pd.concat([list(df.values())[0] for df in dfs])
+    df_all.reset_index(drop=True, inplace=True)
+    return df_all
+
```
abkfenris, line 199: I'd suggest pulling the aggregation of results out into it's own function.

abkfenris (commented):
Nice! It looks a lot cleaner now. 

I think there are still some ways that it could be cleaned up even more that could help the test-ability and reuse-ability. Extracting the data transformation and url generation into their own functions would help that.

For example, the current `fetch_results()` could become two functions that should be come much easier to test individually.

```py
def parse_results(data: bytes) -> Dict[str, DataFrame]:
    df = pd.read_csv(data)
    try:
        df.dropna(subset=[protocol], inplace=True)
    except KeyError:
        return None
    df["Server url"] = url.split("search")[0]
    return {key: df[["Title", "Institution", "Dataset ID", "Server url"]]}


def fetch_results(url: str, key: str, protocol="tabledap") -> Optional[Dict[str, DataFrame]]:
    """
    Parse search results from multiple servers
    """
    data = multi_urlopen(url)
    if data is None:
        return None
    return parse_results(data)
```

Then `parse_results()` can be tested by passing it concrete data, and `fetch_results()` can be tested with `parse_results()` mocked out in addition to using library like VCR.

`parse_results()` could also be then used by a server specific search method on `erddapy.ERDDAP`, or if I wanted to use httpx to make async requests instead. Similarly generating search URL could have multiple uses too. 

It might make sense to pull these functions out into a separate file to keep things more organized.

erddapy/erddapy.py
callumrollo: Implemented cf1d2a8

callumrollo:
Thanks @abkfenris, I've refactored into several smaller functions and pulled all of these out into a separate file. I'll start working on some tests, which should be easier with the more atomic structure you recommended

callumrollo:
Am I on the right track with tests here? I'm not sure how to proceed with the more involved ones like `parse_results` how would we mock data for this?

erddapy/url_handling.py
```diff
@@ -33,6 +33,23 @@ def urlopen(url: str, auth: Optional[tuple] = None, **kwargs: Dict) -> BinaryIO:
     return data
 
 
+def multi_urlopen(url: str) -> BinaryIO:
```
ocefpaf, line 36: Let's "fold" this one into a the canonical `urlopen` by making the latter a thin wrapper to this one. That will allow us to cache the results in that one.

erddapy/url_handling.py
```diff
@@ -33,6 +33,23 @@ def urlopen(url: str, auth: Optional[tuple] = None, **kwargs: Dict) -> BinaryIO:
...
+    """
+    A more simple url open to work with joblib and multiprocessing
```
ocefpaf, line 38:
```suggestion
    A more simple url open to work with multiprocessing
```

erddapy/erddapy.py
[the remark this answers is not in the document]
ocefpaf: @abkfenris thanks for the review and comments! Hope to see you here more often ;-p

erddapy/erddapy.py
ocefpaf: @abkfenris the long term goal would be to use asyncio here. But we are still learning if that makes sense. My draft with joblib was just b/c I'm more familiar with it and local cpu multiprocess is easier for me to understand than asyincio.

callumrollo:
Thanks for the help @abkfenris and @ocefpaf! Can we close #32 now?

PR #201, opened 2021-07-26
deploy docs

ocefpaf:
@callumrollo now I remembered why I've been postponing this for such a long time. We don't have a nice way to publish versioned docs with GitHub Actions yet :-(

Let's leave this here as a draft and I'll work on it this week.


ocefpaf:
@callumrollo the path recommended in https://goerz.github.io/docs_versions_menu/v0.4.1/command.html#deployment-with-github-actions is https://github.com/goerz-testing/cython-package-example/blob/e60098f7d38c4a7a8d2ec73fe6e96016f2b47fa5/.github/workflows/docs.yml

That is way too cumbersome and looks a bit hard to maintain. I'll be looking into alternatives tomorrow. I believe there is another extension that can be used here instead.

PR #203, opened 2021-09-01
Fix multiple search bug

ocefpaf:
The bug was when using the built-in server list, where the namedtuple object was passed instead of the URL string. The rest of the changes are:

- [X] Some functions used here belong only to these functionality and should be private.
- [X] One of those was in another module and, since it will be private, it belongs in the `multiple_server_search.py` instead.
- [ ] Added tests to avoid regression.


ocefpaf:
@callumrollo I forgot to ping you here, sorry! But I do want your feedback on this PR and I'll try to comment out a few points in the code. We can iterate on it again if you have anything you want to change. When you have the time, please take a quick look at my comments.

erddapy/multiple_server_search.py
```diff
@@ -1,47 +1,65 @@
...
+from typing import Dict
+from typing.io import BinaryIO
 
 import pandas as pd
+import requests
 from joblib import Parallel, delayed
 
 from erddapy.servers import servers
-from erddapy.url_handling import format_search_string, multi_urlopen
+from erddapy.url_handling import urlopen
```
ocefpaf, line 10: Made an effort to make `urlopen` the canonical function for both serial and parallel searches. Seems to be working and we have just a thin, private,  wrapper named `_multi_urlopen` now.

erddapy/multiple_server_search.py
```diff
@@ -1,47 +1,65 @@
...
 
 
-def parse_results(data: bytes, protocol, key, url) -> Dict[str, pd.DataFrame]:
+def _format_search_string(server: str, query: str) -> str:
```
ocefpaf, line 13: This one should be here and private. We don't want users to rely on its behavior.

erddapy/multiple_server_search.py
```diff
@@ -1,47 +1,65 @@
...
     """
-    Parse server search results into a pandas DataFrame
+    Generate a search string for an erddap server with user defined query
+    """
+    return f'{server}search/index.csv?page=1&itemsPerPage=100000&searchFor="{query}"'
+
+
+def _format_results(dfs: Dict[str, pd.DataFrame]) -> pd.DataFrame:
+    # we return None for bad server, so we need to filter them here
+    df_all = pd.concat([list(df.values())[0] for df in dfs if df is not None])
```
ocefpaf, line 22: I'm still on the fence about the None vs error here. Upstream the list of servers is smaller and they ensure a valid response. We cast a wider net so iterating over them instead of erroring out is preferred. But users can be surprised if a specific server they hope to see won't show up. Maybe we need to a warning/logging system to catch those.

erddapy/multiple_server_search.py
```diff
@@ -1,47 +1,65 @@
...
+    try:
+        df.dropna(subset=[protocol], inplace=True)
+    except KeyError:
+        return None
+    df["Server url"] = url.split("search")[0]
+    return {key: df[["Title", "Institution", "Dataset ID", "Server url"]]}
 
 
 def search_servers(
-    query="glider",
```
ocefpaf, line 44: This is not a "sane default" but rather a user case. We should make query mandatory arg.

callumrollo:
@ocefpaf thanks for tagging me in this. Looks like you've made great improvements to my rather shoddy first attempt at a multiple search! I'll have a look this evening and try to do a proper code review

erddapy/multiple_server_search.py
callumrollo: I clearly had gliderpy on the brain while implementing this!

erddapy/multiple_server_search.py
callumrollo: A logging system would be good in future. I'd prefer returning None rather than erroring on a failed server response. I got a 503 from coastwatch earlier this week. If the servers we're pulling from don't have near 100 % uptime we don't want to crash queries out.

tests/test_url_handling.py
```diff
@@ -62,25 +56,3 @@ def test_check_url_response():
...
-@pytest.mark.web
-@pytest.mark.vcr()
-def test_multi_urlopen():
-    """Assure that multi_urlopen is always a BytesIO object."""
-    url = "http://erddap.sensors.ioos.us/erddap/tabledap/"
-    ret = multi_urlopen(url)
-    isinstance(ret, io.BytesIO)
-
-
-@pytest.mark.web
```
callumrollo, line 76: Do we need a new test for the refactored multi erddap search? Or has that been implemented somewhere in the new cassettes?

erddapy/multiple_server_search.py
callumrollo: much neater, thanks

callumrollo (commented):
Changes look good to me. Today I learned what private methods are in Python

callumrollo:
Still not 100 % sure on how the GithHub code review works, but I've had a look through the changes and pulled the most recent commits locally. Looks a lot more elegant with the refactoring. I think it's been merged in already? I don't have any major comments, other than we should think of the best way to pass information on empty resonposes/non-responding servers to the downstream user. Not sure what the best way of doing this is

ocefpaf:
> Still not 100 % sure on how the GithHub code review works

Well, I wasn't the best coder here b/c I pinged you after I merged  :grimacing: 
But it is pretty much that. We read the changes and discuss the implications on maintainability, usability, and hopefully how to remove bugs.

> but I've had a look through the changes and pulled the most recent commits locally. Looks a lot more elegant with the refactoring. I think it's been merged in already? I don't have any major comments, other than we should think of the best way to pass information on empty resonposes/non-responding servers to the downstream user. Not sure what the best way of doing this is

We should probably just log/warn on those. Failing is complicated due to the nature of the servers. Otherwise we will end up like Bob's search tool, with a small subset of really "known to work" servers. We can probably add an option to force fail.

PR #210, opened 2021-09-22
Fix iris package name in requirements-dev.txt

pmav99:


requirements-dev.txt
```diff
@@ -9,7 +9,7 @@ flake8-comprehensions
 flake8-mutable
 flake8-print
 interrogate
-iris>=3
+scitools-iris>=3
```
ocefpaf, line 12: This is a shared conda/pip file so we need a better solution. It is quite sad that `iris` has a different name on PyPI. Let me see if I can make a multiple output for the conda version of the package so we can cover both with the PyPI name instead of the canonical name.

ocefpaf: OK. Worked around it in https://github.com/conda-forge/iris-feedstock/pull/74

Issue #212, opened 2021-09-23
relative_constraints is used only in download_url


ocefpaf:
Also, its implementation is odd b/c it requires two sets of constraints instead of one. We should consolidate that and improve the user experience here.

ocefpaf:
Done in #213

PR #213, opened 2021-09-24
better support for relative constraints

ocefpaf:
- [X] download_url
- [x] search_url
- [X] ~~ensure it is only used in tabledap~~ let's let ERDDAP handle failed URLs upstream. Users should read the docs.
- [x] add example/docs


ocefpaf:
This introduces a breaking change to those using relative constraints that was added by @thogar-computer awesome PR in https://github.com/ioos/erddapy/pull/162. However, the fix is easy, just merge all constraints into a single kwargs!

Closes #212

PR #229, opened 2022-01-31
Remove Pandas FutureWarning

vinisalazar:
Hi,

### Problem
When running `ERDDAP.griddap_initialize`, I would get the following warning: `FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.` My Pandas version is 1.4.0.

### Proposed solution
If merged, this PR will modify the `_griddap_get_constraints` function in order to remove the Pandas FutureWarning that was introduced in the latest Pandas versions.

I ran Pytest, Pre-Commit, and tested the function locally (by running `ERDDAP.griddap_initialize`), and things seem to be working properly.

Summary of changes:
* Replace `frame.append()` with `pd.concat()`
* Replace `table.index =` with `.set_index(drop=True)`
* Format code with Black

Thank you for any assistance you can provide,

Vini

### DIFF
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -101,17 +101,16 @@ def _griddap_get_constraints(
             data_start = data[-1][0]
         else:
             data_start = data[0][0]
-        table = table.append(
-            {
+        table = pd.concat([
+            table,
+            pd.DataFrame([{
                 "dimension name": dim,
                 "min": data_start,
                 "max": data[-1][0],
                 "length": len(data),
-            },
-            ignore_index=True,
+            }])],
         )
-    table.index = table["dimension name"]
-    table = table.drop("dimension name", axis=1)
+    table.set_index("dimension name", drop=True)
     constraints_dict = {}
     for dim, data in table.iterrows():
         constraints_dict[f"{dim}>="] = data["min"]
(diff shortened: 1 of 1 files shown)


vinisalazar:
Update: just did some more checks and this introduces some bugs. Will investigate further and push to this PR.

vinisalazar:
Fixed with f94c9b8. Had forgotten to overwrite the new `table` variable.

ocefpaf:
@vinisalazar is this one ready for review?

vinisalazar:
Yes, thank you for the reminder!

erddapy/erddapy.py
```diff
@@ -101,17 +101,22 @@ def _griddap_get_constraints(
...
+                    [
+                        {
+                            "dimension name": dim,
+                            "min": data_start,
+                            "max": data[-1][0],
+                            "length": len(data),
+                        },
+                    ],
+                ),
+            ],
```
ocefpaf, line 117: To avoid this deep nesting we can try probably create the append df before the calling `pd.concat`.

vinisalazar:
@ocefapf apologies for missing your comment about the nesting. Thank you for the review and merge!

ocefpaf:
No problem. I'm doing some housekeeping to get ready for GSoC and went ahead with this one. Hope to see more PRs from you in the future ;-p

PR #232, opened 2022-02-09
refresh cassette

ocefpaf:
Not sure why tests are failing in #231 but let's give this one a try.


ocefpaf:
We cannot use VCR with joblib and we cannot hit the server multiple times. The tests that fall into that category will be run only once for Linux and latest  supported Python.

PR #233, opened 2022-03-14
add project_urls

ocefpaf:
This will improve PyPI navigation.


ocefpaf:
I'll work on the tests in another PR. This one is only about the metadata anyway.

PR #240, opened 2022-03-16
drop py36 and pandas <1

ocefpaf:
At some point we should probably vendor `parse_time_string` b/c that is not part of pandas' public API.


ocefpaf:
Closing and opening to toggle the CIs.

PR #245, opened 2022-04-26
Moving from requests to httpx

ocefpaf:
Most of the time httpx performed slightly better for me and have async, which we plan to use soon.


erddapy/erddapy.py
```diff
@@ -246,7 +242,7 @@ class ERDDAP:
         variables: a list variables to download.
         response: default is HTML.
         constraints: download constraints, default None (opendap-like url)
-        params and requests_kwargs: `request.get` options
+        params and request_kwargs: `httpx.get` options
```
ocefpaf, line 245: @abkfenris this is a breaking change. How do you feel about it? Should we just keep the kw args name `requests`?
In a way the move to `httpx` may break some users in unexpected ways anyway...

abkfenris: How about using a property as an alias and have it emit a deprecation warning?

ocefpaf: Not sure we need to be that complex, at least for the userbase we have at the moment, but I may be wrong.

I was wondering that maybe we should keep the current name for the keyword, renaming it is probably confusing and unnecessary. (I want to move away from requests but we don't need to be that radical, right?)

ocefpaf: Renaming was a brain fart. I needed to read it out loud to realize it. Thanks for bearing with me here. I'll work on fixing the tests next.

ocefpaf:
> [1834b53](https://github.com/ioos/erddapy/pull/245/commits/1834b53b3e3302ce6bfeef9cdb77f4626d6c9754) Not possible. All servers are bad.

Some a less bad then others :-)

At least I can run some live tests with these ones without being kicked out for making two requests within a short period of time.

ocefpaf:
@abkfenris tnis PR is good to go but I'd love your input before merging. The change seems minor, use httpx instead of requests, but the devil is in the details. I hope I did not break anything.

abkfenris (approved):
Looks good to me.

I know the change in exceptions will give me issues in my data pipelines, so I'm guessing it may give others problems also, so its probably worth some good consideration when it's released/what version # it gets. Maybe use it as a basis for the refactoring and call the refactoring and httpx v2?

Issue #96, opened 2019-10-10
query masked by WKT area 


7yl4r:
A common task is how to fetch data masked by an arbitrary region of interest (RoI) polygon.
Ideally the OPeNDAP standard & ERDDAP might allow for submission of a WKT, but masking could also be handled on the client side.

@ocefpaf : what are the first steps towards implementing RoI masking in erddapy?

Additional thoughts: 
1. Should efforts to implement RoI queries be focused on the server or is this client workaround worthwhile?
2. Is this feature creep? Perhaps masking is a secondary task which should be kept separate from the core functionality of the API.

ocefpaf:
We need to implement the `griddap` URL parser first. Working on that. Let's keep this one open until then. Once we have that using any object that has a `_geo_interface_` should be possible.

MathewBiddle:
Possibly a server side function in the future.
xref: https://github.com/BobSimons/erddap/issues/55

PR #246, opened 2022-04-26
area example

ocefpaf:
Waiting on #245.


ocefpaf:
This PR closes https://github.com/ioos/erddapy/issues/96 but we are not implementing a server side slice per se. We are adding a work around using Python software that produces the same results. When that feature is available in ERDDAP itself we should implement a wrapper here too.

Issue #14, opened 2017-11-21
Add allDatasets query method


rsignell-usgs:
In my ERDDAP time series explorer, I used the existing erddapy method for advanced search , but also needed to use an allDatasets query to return the lon/lat values of all stations.   I did this with an ugly erddap URL string formatting:
https://github.com/reproducible-notebooks/ERDDAP_timeseries_explorer/blob/master/ERDDAP_timeseries_explorer.py#L197-L200

Seems like the allDatasets query might warrant a method, since it's a bit different than a regular datasets query

rsignell-usgs:
I just realized we don't need a special `allDatasets` query method, because `allDatasets` is just a special dataset, so we can use something like:
```
    kwargs = {
        'minTime<=': stop_time,
        'maxTime>=': start_time,
    }

    variables = ['datasetID', 'minLongitude', 'minLatitude']
    url = e.get_download_url(dataset_id='allDatasets', variables=variables, response='csv', **kwargs)
```
to generate ERDDAP URLs like this one, which returns the `ID`, `min_lon` and `min_lat` for all datasets in a certain time range

https://erddap-uncabled.oceanobservatories.org/uncabled/erddap/tabledap/allDatasets.csv?datasetID,minLongitude,minLatitude&minTime%3C=2018-01-23T12:52:10Z&maxTime%3E=2018-01-16T12:52:10Z

ocefpaf:
>  url = e.get_download_url(dataset_id='allDatasets', variables=variables, response='csv', **kwargs)

We should document that.

ocefpaf:
I forgot to document this. Re-opening so we can do that.

ocefpaf:
The docs were added in https://github.com/vinisalazar/erddapy/commit/f0ec8a64aed995137495a24bd9414e53ea9d21a4. Closing this now.

PR #251, opened 2022-06-13
Document search using 'allDatasets' method

vinisalazar:
Hi, this is in relation to #14.

**Summary of changes**
  - Add cell to '00-quick_intro.ipynb' documenting this type of search

The notebook built as expected using the command in the `deploy-docs.yaml` file.

Please feel free to request any changes.

Thank you,
Vini


ocefpaf:
> I believe I'm running into the same problem as [described here](https://github.com/deathbeds/prenotebook/issues/6), pre-commit passes locally but fails on the CI.

We should disable that pre-commit check. It is causing more harm than good at this porint. Do you mind sending another PR to do that?

Issue #244, opened 2022-04-21
Remove pagination from all responses but HTML in search_for


ocefpaf:
Items per page does not make sense and can be misleading in non-HTML responses. When someone requests a JSON or CSV they want all the results.

ocefpaf:
@vinisalazar this is probably the last issue you can address before we lay out a plan for the refactor.

ocefpaf:
@vinisalazar if I recall correctly you addressed this one, right? Can we close it?

vinisalazar:
Yep!

PR #255, opened 2022-06-20
Increase number per items on CSV, JSON searches

vinisalazar:
Hi,

this is in relation to #244. I wasn't sure of the best way to go about this, so I tried keeping it as simple as possible. Please let me know if there's an obvious way that I'm missing.

What I did was simply to set the number of `items_per_page` to a big number (initially it was `1e9`, but I thought that was excessive?) if the response is either `'csv'` or `'json'`. This doesn't require actually making the request, it just builds the URL. I couldn't find a "number of records" variable in the search response. I suppose I could set the `items_per_page` variable to that, but that would require making the request, and I understand that we just want the URL. Removing the `items_per_page` or `page` variables from the URL didn't make any difference.

To test, I added a fixture for the gliders server, since the other two had less than 1000 records. I did not commit the cassettes since they turned out fairly large, at around 3MB each. Do you have any suggestions on what to do about this?

**Summary of changes**
- Modify variable `items_per_page` if `response` is either `'csv'` or `'json'`
- Add new fixture `gliders` to `test_to_objects.py`
- Add tests `test_csv_search` and `test_json_search`

Please feel free to request any changes, and don't forget to add the `gsoc-2022` label :)

Thank you,
Vini

### DIFF
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -74,8 +74,7 @@ def _griddap_get_constraints(
     """
     Fetch metadata of griddap dataset and set initial constraints.
 
-    Step size is applied to all dimensions
-
+    Step size is applied to all dimensions.
     """
     dds_url = f"{dataset_url}.dds"
     url = urlopen(dds_url)
@@ -399,8 +398,10 @@ class ERDDAP:
         protocol = protocol if protocol else self.protocol
         response = response if response else self.response
 
-        # Different protocol if not dealing with pagination, return all records.
-        # This is done for all CSV, JSON and TSV-style responses.
+        # These responses should not be paginated b/c that hinders the correct amount of data silently
+        # and can surprise users when the number of items is greater than ERDDAP's defaults (1000 items).
+        # Ideally there should be no pagination for this on the ERDDAP side but for now we settled for a
+        # "really big" `items_per_page` number.
         non_paginated_responses = [
             "csv",
             "csvp",
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -385,7 +385,7 @@ class ERDDAP:
 
             response: default is HTML.
             items_per_page: how many items per page in the return,
-                default is 1000.
+                default is 1000 for HTML, 1e6 (hopefully all items) for CSV, JSON.
             page: which page to display, default is the first page (1).
             kwargs: extra search constraints based on metadata and/or coordinates ke/value.
                 metadata: `cdm_data_type`, `institution`, `ioos_category`,
@@ -398,6 +398,12 @@ class ERDDAP:
         """
         protocol = protocol if protocol else self.protocol
         response = response if response else self.response
+
+        # Different protocol if not dealing with pagination, return all records
+        non_paginated_responses = ["csv", "json"]
+        if response in non_paginated_responses:
+            items_per_page = int(1e6)
+
         return _search_url(
             self.server,
             response=response,
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -399,8 +399,20 @@ class ERDDAP:
         protocol = protocol if protocol else self.protocol
         response = response if response else self.response
 
-        # Different protocol if not dealing with pagination, return all records
-        non_paginated_responses = ["csv", "json"]
+        # Different protocol if not dealing with pagination, return all records.
+        # This is done for all CSV, JSON and TSV-style responses.
+        non_paginated_responses = [
+            "csv",
+            "csvp",
+            "csv0",
+            "json",
+            "jsonlCSV1",
+            "jsonlCSV",
+            "jsonlKVP",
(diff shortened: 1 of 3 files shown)


vinisalazar:
I think that the test failing here may be fixed by #253.

abkfenris:
While you could write a mock response generator, you then would be testing that more so than `.get_search_url()`. 

In this case, I think we can skip calling the endpoints, and instead inspect the URL returned from `.get_search_url()` directly to make sure that it is the formatting differently based on the response type.

Additionally we could leave these tests in in addition to testing the URL, mark them, and have pytest skip them by default. Then maybe we set up a weekly cron driven actions workflow to run those tests and update the VCR cassettes to do a fuller test run?

vinisalazar:
Hi @abkfenris, sounds good to me, I'll add new tests that check the URL. However, I'm still thinking we should drop the test on the CoastWatch server? That seems to be behaving inconsistently. Please let me know what you think works best.

Thanks,
Vini

abkfenris:
Ya, avoiding the Coastwatxh server might be best with Bobs aggressive blocklist.

vinisalazar:
Do you think you could merge #253 then? Please let me know if you'd like me to make any changes to that.

Thanks,
Vini

erddapy/erddapy.py
```diff
@@ -385,7 +385,7 @@ def get_search_url(
 
             response: default is HTML.
             items_per_page: how many items per page in the return,
-                default is 1000.
+                default is 1000 for HTML, 1e6 (hopefully all items) for CSV, JSON.
```
ocefpaf, line 388: I was hoping that no pagination would return everything but it seems that ERDDAP appends the pagination even in `csv`, etc.

vinisalazar: Me too, I tried setting it to zero but it didn't work, I'm wondering there are any more elegant workarounds or should we just leave it as a large number?

erddapy/erddapy.py
```diff
@@ -398,6 +398,12 @@ def get_search_url(
         """
         protocol = protocol if protocol else self.protocol
         response = response if response else self.response
+
+        # Different protocol if not dealing with pagination, return all records
+        non_paginated_responses = ["csv", "json"]
```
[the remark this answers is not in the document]
vinisalazar: Sorry, this is not very clear to me. Don't all of the response types append pagination? And we would like to keep pagination for some (e.g. HTML), and not for others (like CSV and JSON). Are there any criteria as to which responses to include in this list? For now I am thinking to include the other CSV, JSON and TSV "variants".

erddapy/erddapy.py
ocefpaf: Let's leave a higher number for now. We can ask Bob someday about it.

erddapy/erddapy.py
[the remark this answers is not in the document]
vinisalazar: I added all CSV, JSON and TSV responses for now.

ocefpaf:
@vinisalazar do you mind rebasing this one to fix the conflict?

Please let me know if you need help with that!

vinisalazar:
@ocefpaf I had a go at doing it, not sure if I got it right. I can reset the branch and redo it if needed.

vinisalazar:
I'll definitely accept the offer for helping out with this one!

ocefpaf:
Looks like you got it. At least the original changes are in there, only one tests is failing.

tests/test_to_objects.py
```diff
@@ -64,6 +75,26 @@ def dataset_tabledap(sensors):
...
+    url = gliders.get_search_url(search_for="all", response="csv")
+    handle = httpx.get(url)
+    nrows = len(list(handle.iter_lines())) - 1
+    assert nrows > 1000
+
+
+@pytest.mark.web
+def test_json_search(gliders):
+    """Test if a JSON search returns all items (instead of the first 1000)."""
+    # The gliders server has 1244 datasets at time of writing
```
ocefpaf, line 91: This comment can go inside the gliders fixture instead of both tests.

erddapy/erddapy.py
```diff
@@ -398,6 +398,24 @@ def get_search_url(
         """
         protocol = protocol if protocol else self.protocol
         response = response if response else self.response
+
+        # Different protocol if not dealing with pagination, return all records.
+        # This is done for all CSV, JSON and TSV-style responses.
```
ocefpaf, line 403: Let's rephrase this for our future self:

```python
# These ERDDAP responses should not be paginated b/c that hinders the correct amount of data silently
# and can surprise users when the number of items is greater than ERDDAP's defaults (1000 items).
# Ideally there should be no pagination for this on the ERDDAP side but for now we settled for a
# "really big" `items_per_page` number.

vinisalazar:
Thank you for the review @ocefpaf!

Issue #258, opened 2022-07-14
Move tests that hit a server to the ERDDAP gold standard server


ocefpaf:
Let's try to move all of our test to use these two servers:

- https://standards.sensors.ioos.us/erddap/index.html
- http://erddap.ioos.us/erddap/index.html

cc @MathewBiddle

MathewBiddle:
Let me know if there are special configurations you need for erddap.ioos.us and we can set it up.

vinisalazar:
Hi, I'm working on moving tests to these two servers, but I don't think there are any griddap datasets on either of them. Would it be possible to add some griddap datasets or should I keep griddap tests as they are (they currently use the [neracoos](http://www.neracoos.org/erddap/index.html) server)?

Thanks!

MathewBiddle:
Might be able to use [EDDGridFromERDDAP](https://coastwatch.pfeg.noaa.gov/erddap/download/setupDatasetsXml.html#EDDGridFromErddap) to point to a specific dataset on NERACOOS. I'll have to do some digging to see how the configuration should be done.

abkfenris:
EDDGridFromERDDAP will still redirect any requests for data along to NERACOOS, but it will handle metadata locally.

MathewBiddle:
Since creating an EDDGridFromERDDAP wont help here. What can we do? Is there another dataset we can host on erddap.ioos.us to run these tests against?

MathewBiddle:
Are any of the datasets in the ioos-gold-standard applicable? 

https://github.com/ioos/erddap-gold-standard/tree/master/datasets

if not, maybe we should add a gridded example dataset there?

ocefpaf:
> if not, maybe we should add a gridded example dataset there?

For now the tests are passing. Thanks to some of the server changes and/or the pytest-vcr recording. The docs are failing though b/c GitHub Actions was blocked from accessing those servers :-/

I'll re-write the docs today and let you know how that goes.

ocefpaf:
Let's revisit this when we have a griddap dataset in the server. For now we are OK and the docs are building again.

Issue #257, opened 2022-07-07
IOOS RA query/utility package


MathewBiddle:
Similar to [gliderpy](https://github.com/ioos/gliderpy) and various other wrapper packages, we should think about an IOOS wrapper package for high level queries against the various IOOS ERDDAP endpoints. Maybe provide some high level utilities too?

A subset of RA endpoints are documented here [1]. However, it would probably be better to use the awesome-erddap json instead [2]. Or, identify another curated list.

[1] https://github.com/ioos/ioos-asset-inventory/blob/main/utils/ra_erddaps.json

[2] https://github.com/IrishMarineInstitute/awesome-erddap/blob/master/erddaps.json

ocefpaf:
> I think we would probably want to wait until after the GSOC refactoring to work on this, but maybe it's a good code sprint project.

Indeed. However, we can start discussion this high level API now. For example, would something like this make sense?

```python
from ioos_erddap import ERDDAP

ds = ERDDAP(ra="secoora")
ds   ← list all dataset_ids

ds["dataset_id"] ← return a xarray object for that dataset using the OPeNDAP response to take advantage of the lazy loading.
```

and

```python
ds = ERDDAP(ra="secoora", **constraints)
ds   ← list all dataset_ids constrained in space/time/query string

ds["dataset_id"] ← return an in memory/fsspec xarray object for the constrained dataset
```

ocefpaf:
@MathewBiddle do you believe it is worth adding this as a topic for the code sprint this year?

MathewBiddle:
Yes!

Harish-2003:
may i know what the code sprint is all about? and is this issue still open...?

ocefpaf:
Hi @Harish-2003, you may find more information on the code sprint at https://ioos.github.io/ioos-code-sprint/

The short version is: The IOOS code sprint is where some of us sit together to try to solve a collection of selected issues and/or add features to our Software stack.

mwengren:
Could we use the existing code base of [ioos/colocate](https://github.com/ioos/colocate) as a starting point?

ocefpaf:
> Could we use the existing code base of [ioos/colocate](https://github.com/ioos/colocate) as a starting point?

This would be lower level. The goal would be for colocate to be a front-end for it.

PR #259, opened 2022-07-18
Start refactoring project

vinisalazar:
Hi,

I have started work on the refactoring project. This PR creates the `core` subpackage and moves the `url_handling` and `netcdf_handling` modules to that. I was thinking of first consolidating the module/directory structure and then moving on to making the code changes.

I was unsure of whether the `erddapy` module should go inside the `core` subpackage. From my understanding, we are eventually going to get rid of the `ERDDAP` class in favour of the new classes in the object layer (or maybe keep it for backwards-compatibility?).  Should other functions in that module (e.g. `_search_url`, `parse_dates`, etc.) be moved into a separate module of the `core` subpackage?

The same goes for the server modules (`servers.py`, `multiple_server_search.py` and the `erddaps.json` file). Should those go into `core` or into a subpackage of their own?

Lastly, should we create a `refactor` or `layer-refactoring` branch to merge this PR into?

Please let me know what you think.

Thank you,
Vini

**Summary of changes**
- Add files to .gitignore
- Create `core` subpackage
- Move `url_handling` and `netcdf_handling` to new subpackage
- Refactor imports


abkfenris (commented):
> I have started work on the refactoring project. This PR creates the `core` subpackage and moves the `url_handling` and `netcdf_handling` modules to that. I was thinking of first consolidating the module/directory structure and then moving on to making the code changes.
> 
> I was unsure of whether the `erddapy` module should go inside the `core` subpackage. From my understanding, we are eventually going to get rid of the `ERDDAP` class in favour of the new classes in the object layer (or maybe keep it for backwards-compatibility?). Should other functions in that module (e.g. `_search_url`, `parse_dates`, etc.) be moved into a separate module of the `core` subpackage?

I think we should keep the existing `ERDDAP` class and the `erddapy` module at their current locations for backwards compatibility (eventually with some deprecation warnings). Most of the functions and methods in there should become wrappers around various things in `core`.

> The same goes for the server modules (`servers.py`, `multiple_server_search.py` and the `erddaps.json` file). Should those go into `core` or into a subpackage of their own?

Hmm, good question. I think we'll probably end up with new, more opinionated multiple server search for the object layer, but the result wrangling will probably be shared. Maybe `core/search_result.py` when those guts get refactored out, with the existing `multiple_server_search.py` left in place for compatibility. I'll have to ponder more on `servers.py` and `erddaps.json` since those are small but combine fetching and wrangling, but are still really helper functions, unless @ocefpaf has some ideas.
 
> Lastly, should we create a `refactor` or `layer-refactoring` branch to merge this PR into?

A `refactor` branch should do the trick.

vinisalazar:
Thank you for your comments @abkfenris. I noticed that checks passed now after [failing on the multiple server search](https://github.com/ioos/erddapy/actions/runs/2691500233). I will work on #258 along with the refactoring and hopefully that will prevent randomly failed tests in the future.

ocefpaf:
>  I'll have to ponder more on servers.py and erddaps.json since those are small but combine fetching and wrangling, but are still really helper functions, unless @ocefpaf has some ideas.

As we discussed in our meeting that is mostly to check the awesome erddap json server list and create the shortcuts to the servers. We will create a package based on the RAs, see #257, and having those into their own sub-module will be useful in the near future.

vinisalazar:
After force-pushing to fix a line-break, the test failed again, so I'm thinking it's probably best to work on #258 before merging this one (I will mark it as a draft until then).

vinisalazar:
I'm hoping that the failing check will pass after merging #260.

vinisalazar:
I added more changes to this branch, but if it's too much, I can reset it and wait for it to be merged before submitting a separate PR.

**Summary of changes**
- Move `doc_helpers.py` module to `notebooks`
- Move helper functions from `multiple_server_search.py` to `url_handling.py` (except for `_format_results` to avoid pandas import in the URL module)
- Create `servers` directory

vinisalazar:
There is a certain degree of repetition for an explicit import of the `servers` dictionary (`from erddapy.servers.servers import servers`) but it can also be imported from the top-level package: `from erddapy import servers`.

vinisalazar:
I just realised that users doing `from erddapy.servers import servers` (as would be done before merging this) will import the module and not the dictionary. Maybe there should be a warning for that?

vinisalazar:
Would it be better to create the `refactor` branch to merge this PR into? Or is `main` ok?

vinisalazar:
I'm not sure why the test for multiple server search fails (apparently) randomly. Should we replace it with a VCR response?

ocefpaf:
> I'm not sure why the test for multiple server search fails (apparently) randomly. Should we replace it with a VCR response?

I could not make VRCs work for that test. I don't remember why. Maybe it was the pytest parameters? They may work if we split them up.

> Would it be better to create the `refactor` branch to merge this PR into? Or is `main` ok?

I prefer main to avoid a huge merge from a refactor branch into main later.

abkfenris:
I'm guessing with multi-server search VCR was probably having a hard time matching request to response. It might need a [custom request matcher](https://vcrpy.readthedocs.io/en/latest/advanced.html#register-your-own-request-matcher)

PR #260, opened 2022-07-27
Refactor tests to use gold standard servers

vinisalazar:
Hi,

this PR is in relation to #258.

I refactored tests to use the servers listed on the issue. I kept some tests as they are (in `test_to_objects.py`) and modified `test_search_url_change_protocol` in `test_url_builder.py` while there aren't any griddap datasets.

While working on `test_to_objects.py`, I ran into a problem with the `to_xarray` method. I wasn't able to read the NC dataset from memory, and detailed it in #261.

Thanks,
Vini

**Summary of changes**
- Refactor tests to use gold standard ERDDAP servers
- Modify `dataset_id`, `constraints`, and `variables`
- Rewrite cassettes


ocefpaf (approved):

ocefpaf:
@vinisalazar the vrc cassetes are quite noisy in the PRs. if we can, when updating them, let's try to commit squash all the previous commits and add them after so we have only 2 commits: code changes and cassette updates.

vinisalazar:
Sure, that makes sense, thanks for the feedback.

PR #264, opened 2022-08-11
Refactor url methods

vinisalazar:
Hi,

this is a downstream branch of #263, so it will be easier to review after merging that and #259. It includes two additional commits that move `get_search_url`, `get_info_url` and `get_categorize_url` from the `erddapy.erddapy` module to the `erddapy.core.url` module.

For `get_download_url`, it's a bit more tricky: it would incur in a circular import between `erddapy.core.url` and the newly-added `erddapy.core.griddap` module, because of the `urlopen` and `_urlopen` functions. So, I'm thinking whether I should move those functions to a new module (maybe `erddapy.core.helpers` or `erddapy.core.__init__`?). Please let me know what you think is best.

Thank you,
Vini

**Summary of changes** (in relation to #263):
- Move the `get_search_url`, `get_info_url`, and `get_categorize_url` to the URL module
- Refactor the private `_search_url` function as public
- Edit docstrings to use gold standard servers

### DIFF
--- a/erddapy/core/url.py
+++ b/erddapy/core/url.py
@@ -11,6 +11,8 @@ import httpx
 import pytz
 from pandas._libs.tslibs.parsing import parse_time_string
 
+OptionalStr = Optional[str]
+
 
 @functools.lru_cache(maxsize=256)
 def _urlopen(url: str, auth: Optional[tuple] = None, **kwargs: Dict) -> BinaryIO:
@@ -59,7 +61,7 @@ def _distinct(url: str, **kwargs: Dict) -> str:
     For example, a query for the variables ["stationType", "stationID"] with `distinct=True`
     will return a sorted list of "stationIDs" associated with each "stationType".
 
-    See https://coastwatch.pfeg.noaa.gov/erddap/tabledap/documentation.html#distinct
+    See http://erddap.ioos.us/erddap/tabledap/documentation.html#distinct
 
     """
     distinct = kwargs.pop("distinct", False)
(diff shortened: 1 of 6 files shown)


vinisalazar:
Added [`0484fbe`](https://github.com/ioos/erddapy/pull/264/commits/0484fbeea2ba78a9899f8e7711839a3b53cc05a4) which refactors the `get_download_url` method.

erddapy/erddapy.py
```diff
@@ -13,12 +12,10 @@
 )
 from erddapy.core.netcdf import _nc_dataset, _tempnc
 from erddapy.core.url import (
-    _check_substrings,
-    _distinct,
-    _format_constraints_url,
-    _quote_string_constraints,
-    _search_url,
-    parse_dates,
```
vinisalazar, line 21: I considered not removing these imports but pre-commit is failing because of that.

vinisalazar: This would spare the need of refactoring imports in argopy, e.g. [this line](https://github.com/euroargodev/argopy/blob/0c408e9f8fc15e7ae68802703b66b1768345a9d9/argopy/data_fetchers/erddap_index.py#L34)

vinisalazar: I added [`04310c7`](https://github.com/ioos/erddapy/pull/264/commits/04310c7b77f057e99a6b159dcdef3d61a96c722b) with an `__all__` variable to include these imports in the module.

vinisalazar:
Added [`66545d1`](https://github.com/ioos/erddapy/pull/264/commits/66545d14544593665bcfa3ccfedbaaf143c86da9) in an effort to prevent tests failing due to server problems, but not sure if a specific formatting was needed for that file.

tests/test_netcdf_handling.py
```diff
@@ -15,7 +15,7 @@ def test__nc_dataset_in_memory_https():
     """Test loading a netcdf dataset in-memory."""
     from netCDF4 import Dataset
 
-    url = "https://podaac-opendap.jpl.nasa.gov/opendap/allData/modis/L3/aqua/11um/v2019.0/4km/daily/2017/365/AQUA_MODIS.20171231.L3m.DAY.NSST.sst.4km.nc"  # noqa
+    url = "http://erddap.ioos.us/erddap/tabledap/allDatasets.nc"  # noqa
```
ocefpaf, line 18: That is probably better than a specific dataset_id. Thanks!

erddapy/core/url.py
```diff
@@ -59,7 +63,7 @@ def _distinct(url: str, **kwargs: Dict) -> str:
     For example, a query for the variables ["stationType", "stationID"] with `distinct=True`
     will return a sorted list of "stationIDs" associated with each "stationType".
 
-    See https://coastwatch.pfeg.noaa.gov/erddap/tabledap/documentation.html#distinct
+    See http://erddap.ioos.us/erddap/tabledap/documentation.html#distinct
```
ocefpaf, line 66: @MathewBiddle we should probably move this to `https` at some point.

Issue #228, opened 2022-01-26
GSoC 2022 ideas


ocefpaf:
Some users are looking for tools to help them assemble ERDDAP urls for use in their own workflows, while others would prefer to work at a higher, more opinionated level. I believe we can more cleanly separate functionality to help support the spectrum of erddapy users better.

Originally errdapy was meant to be a url builder only. We added the main class later and stop in between trying to support many different usage patterns via the single primary class [`ERDDAP`](https://github.com/ioos/erddapy/blob/9653982fbfabbf6b9310ec356bcfa65519564291/erddapy/erddapy.py#L68).

## Issues to address

- Users in interactive workflows are required to transform `ERDDAP` objects as they wish to connect to new datasets, for example when moving from searching a server to visiting the datasets.
- Adding constraints to a `ERDDAP` object are stateful in place changes, where as most interactive users are used to Numpy/Pandas/xarray style workflows where you can return or chain together changes.
- Switching out IO is currently non-trivial due to URL generation and data transformations being tightly coupled to IO.

## Proposed Solution

I proposed that we separate erddapy into more functional layers, roughly following the [SQLAlchemy core/ORM](https://docs.sqlalchemy.org/en/14/intro.html#overview) model.

### Core Layer

The core layer would contain two primary components of functionality: url generation & data transformation. This layer makes no choices or assumptions about IO allowing it to be reused easily.

- URL generation - Functions to generate valid URLs from bare components, such as a dataset name, format, and dictionary of constraints to `tabledap/M01_met_all.csv?time%2Cair_temperature%26air_temperature_qc%3D0%26time>%3D"2020-12-09T15%3A25%3A00.000Z"`
- Data transformation - Functions to convert a raw response (.csv, .nc, ...) into Pandas DataFrames, xarray Datasets...

### Object (or opinionated) Layer

The object (or opinionated) layer would present higher level objects for searching servers and accessing datasets with a Pandas or xarray like returning or chainable API compared to the transformational API of the current `ERDDAP` class. 
This layer uses much of the core functionality and presents it in easy to use ways with an opinion as to the access method.

Additionally if possible these objects should be serializable, so they can be pickled and passed to other processes/machines (Dask/Dagster/Prefect).

- `class ERDDAPConnection`
    - While most ERDDAP servers allow connections via a bare url, some servers may require authentication to access data.
    - `.get(url_part: str) -> bytes or str`
        - Method actually request data.
[...]

Issue #261, opened 2022-07-27
`to_xarray` method error on OS X


vinisalazar:
Hi,

As described in #260, I am getting an error when trying to use the `to_xarray` method on OS X. Here is the code to reproduce the error:
```python
In [1]: from erddapy import ERDDAP

In [2]: e = ERDDAP(
   ...:         server="https://standards.sensors.ioos.us/erddap/",
   ...:         response="htmlTable",
   ...:     )

In [3]: e.dataset_id = "org_cormp_cap2"

In [4]: e.protocol = "tabledap"

In [5]: e.variables = ["sea_water_temperature", "time"]

In [6]: e.constraints = {
   ...:         "time>=": "1999-10-01T00:00:00Z",
   ...:         "time<=": "1999-10-01T21:00:00Z",
   ...:         "latitude>=": 30,
   ...:         "latitude<=": 33,
   ...:         "longitude>=": -80,
   ...:         "longitude<=": -78,
   ...:     }
```

`to_pandas()` and `to_iris()` work (although I get some UserWarnings on the latter, not sure if they are expected):

```python
In [7]: e.to_pandas()
Out[7]: 
    sea_water_temperature (degree_Celsius)            time (UTC)
0                                    27.30  1999-10-01T00:08:00Z
1                                    27.30  1999-10-01T01:08:00Z
2                                    27.26  1999-10-01T02:08:00Z
3                                    27.24  1999-10-01T03:08:00Z
4                                    27.27  1999-10-01T04:08:00Z
5                                    27.32  1999-10-01T05:08:00Z
6                                    27.34  1999-10-01T06:08:00Z
7                                    27.35  1999-10-01T07:08:00Z
8                                    27.36  1999-10-01T08:08:00Z
9                                    27.33  1999-10-01T09:08:00Z
10                                   27.30  1999-10-01T10:08:00Z
11                                   27.31  1999-10-01T11:08:00Z
12                                   27.30  1999-10-01T12:08:00Z
13                                   27.28  1999-10-01T13:08:00Z
14                                   27.29  1999-10-01T14:08:00Z
15                                   27.31  1999-10-01T15:08:00Z
16                                   27.37  1999-10-01T16:08:00Z
17                                   27.42  1999-10-01T17:08:00Z
18                                   27.49  1999-10-01T18:08:00Z
19                                   27.56  1999-10-01T19:08:00Z
20                                   27.60  1999-10-01T20:08:00Z

In [8]: e.to_iris()
[...]

PR #267, opened 2022-09-13
Draft classes of object/opinionated layer

vinisalazar:
Hi,

this PR is a draft of how the new object/opinionated layer should look like. It does not have any "proper" code yet, just a scaffold of how the classes would look like. I am submitting it so we can discuss the best way of implementing this layer based on the discussion on #228.

I have a couple of questions:
- What would be the best name for the subpackage and main module? I went with `objects/` and `objects.py` so there wouldn't be any confusion with the Python `object` reserved word, but I also considered `orm` as the name.
- On #228, @ocefpaf suggests that on the `ERDDAPConnection` class:
> `@property(server) -> ERDDAPConnection`
> Return a new ERDDAPConnection if trying to set a new server, or change other attributes rather than changing it in place.

I wasn't sure of what's the best way to do this. I wrote a `.to_string` method which takes either a string (with the new server URL) or an instance of `ERDDAPConnection` and returns just a string, or a `TypeError`:

https://github.com/vinisalazar/erddapy/blob/257de3d4144c6a7011d36e0617aa1b89ebf79287/erddapy/objects/objects.py#L22-L33

That function is then used in the `.server` setter and in the `ERDDAPServer` `__init__` and `connection`. Should the `ERDDAPConnection.server` setter return a new instance? E.g.

```python
    @connection.setter
    def connection(self, value):
        return ERDDAPConnection(value)
``` 

Once you approve this, I will start implementing the `core` functions in this layer. Looking forward to hearing your feedback.

Thanks,
Vini


vinisalazar:
I think `test_to_iris_tabledap` is failing because of the latest version of iris. I can reproduce the same error after updating to `3.3.0`.

I'll attempt a fix on a separate PR.

erddapy/objects/objects.py
```diff
@@ -0,0 +1,189 @@
...
+
+    pass
+
+
+class ERDDAPServer:
+    """Instance of an ERDDAP server, with support to ERDDAP's native functionalities."""
+
+    def __init__(self, connection: str | ERDDAPConnection):
+        """Initialize instance of ERDDAPServer."""
+        self._connection = ERDDAPConnection(ERDDAPConnection.to_string(connection))
```
abkfenris, line 163:
```suggestion
    def __init__(self, url: str, connection: ERDDAPConnection | None):
        """Initialize instance of ERDDAPServer."""
        if "http" in url:
            self.url = url
        else:
            # get URL from dict of ERDDAP servers
        self._connection = connection or ERDDAPConnection()
```

If the connection doesn't include server info, then the connection can be simplified.

We could also support initializing servers based on whats in the [`servers.server_list`](https://github.com/ioos/erddapy/blob/6057872825aa29423e31c849a3c921fe25474e21/erddapy/servers/servers.py#L10) as we currently do.

erddapy/objects/objects.py
```diff
@@ -0,0 +1,189 @@
...
+        connection: str | ERDDAPConnection,
+        variables,
+        constraints,
+    ):
+        """Initialize instance of ERDDAPDataset."""
+        self.dataset_id = dataset_id
+        self._connection = ERDDAPConnection(ERDDAPConnection.to_string(connection))
+        self._variables = variables
+        self._constraints = constraints
+        self._meta = None
```
abkfenris, line 78: For the datasets, I envisions supporting some subset of the xarray/pandas filtering/subsetting API.

For instance if I had a `GridDataset` for http://www.neracoos.org/erddap/griddap/WW3_72_GulfOfMaine_latest.html as `eds` I could do something like to filter down:

```py
eds = GridDataset.from_url("http://www.neracoos.org/erddap/griddap/WW3_72_GulfOfMaine_latest.html")

eds = eds[['hs', 'dir']]  # select just the wave height and direction variables
eds = eds.sel(time=slice("2022-09-23T00:00:00Z", "2022-09-26T00:00:00Z")) # only for a subset of the time range
```

Then there are different ways that we could consume the dataset.

```py
ds = eds.to_xarray() # creates a xarray dataset via OpenDAP and filters/slices based on the earlier operations

netcdf_url = eds.url('nc')  # return the NetCDF url

eds.save('nc', NETCDF_PATH)  # Save a netcdf locally.
```

We probably want to validate that filtering/slicing is working on legit values by default, but have a way to enable/disable it both globally/per dataset. To validate it, we would need to retrieve all the valid variables and dimensions from the server.

abkfenris (commented):
One thing that would probably help is to break up the classes into separate files. As is, maybe read my notes starting with connection, then server, then datasets.

>     * What would be the best name for the subpackage and main module? I went with `objects/` and `objects.py` so there wouldn't be any confusion with the Python `object` reserved word, but I also considered `orm` as the name.

What about something like `array_like`? 

Even if we aren't implementing any/all of the Python array manipulation dunder-methods at this point, it gives us an interface something to aspire to.

>     * On [GSoC 2022 ideas #228](https://github.com/ioos/erddapy/issues/228), @ocefpaf suggests that on the `ERDDAPConnection` class:
> 
> 
> > `@property(server) -> ERDDAPConnection`
> > Return a new ERDDAPConnection if trying to set a new server, or change other attributes rather than changing it in place.
> 
> I wasn't sure of what's the best way to do this. I wrote a `.to_string` method which takes either a string (with the new server URL) or an instance of `ERDDAPConnection` and returns just a string, or a `TypeError`:
> 
> https://github.com/ioos/erddapy/blob/257de3d4144c6a7011d36e0617aa1b89ebf79287/erddapy/objects/objects.py#L22-L33
> 
> That function is then used in the `.server` setter and in the `ERDDAPServer` `__init__` and `connection`. How should the `ERDDAPConnection.server` setter return a new instance? This doesn't look like it would work:
> 
> ```python
>     @connection.setter
>     def connection(self, value: str):
>         return ERDDAPConnection(value)
> ```

See my inline comments, but I think we can simplify the connection and servers so that this isn't an issue.

vinisalazar:
Hi @ocefpaf and @abkfenris,

thank you very much for the review. I just arrived in Australia after a long trip and will take a couple of days to settle, so I will address the comments next week, and for this week I will attempt a fix to #261. Please let me know if there are any problems.

Best,
V

vinisalazar:
The dataset used in `test_to_objects.py` is blocking the CI from passing. It's necessary to merge #271 or #272 before continuing

vinisalazar:
Closing as stale

PR #268, opened 2022-09-13
Improve exception handling in method to_iris

vinisalazar:
Hi,

this is a fix to the problem mentioned in #267.

On the `ERDDAP.to_iris()` method, when the `cubes.realise_data()` method doesn't work, it realises the data by calling `cube.data`. The new versions of Iris prevent creating a `CubeList` object using `cube.data` objects (as they are Numpy Arrays and not cubes), so that behaviour was modified.

Summary of changes:
  - Realise data by creating placeholder variable `_`
  - Don't use `iris.cube.CubeList` for placeholder variable

Thanks,
Vini

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ dependencies:
   - python=3.9
   - cartopy
   - erddapy
-  - iris
+  - iris>=3.3.0
   - jupyter
   - pendulum>=2.0.1
   - xarray
--- a/environment.yml
+++ b/environment.yml
@@ -7,5 +7,6 @@ dependencies:
   - erddapy
   - iris>=3.3.0
   - jupyter
+  - netcdf4<1.6.1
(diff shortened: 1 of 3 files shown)


ocefpaf:
@vinisalazar let's just constrain our minimum iris version to the latest so we can remove the rest of the try-except.

vinisalazar:
Thank you for the review @ocefpaf 

I added the minimum version to `environment.yaml`, are there any other files that must be updated? e.g. `requirements.txt`, or even the conda-forge feedstock?

ocefpaf:
Can you try pinning netcdf4<1.6.1. I believe we are hitting https://github.com/Unidata/netcdf4-python/issues/1192

vinisalazar:
I force-pushed a new commit that adds the netcdf4 pin to `requirements-dev.txt` as well.

PR #273, opened 2022-10-13
Update actions version in all GHA workflows

douglasnehme:
Change all actions version to the newest tagged ones, including which were pointing to the main branch.


.github/workflows/deploy-docs.yml
```diff
@@ -11,12 +11,12 @@ jobs:
 
     steps:
     - name: checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3.1.0
```
ocefpaf, line 14: We can use all "3.*" instead of a point release.

```suggestion
      uses: actions/checkout@v3
```

.github/workflows/deploy-docs.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         popd
 
     - name: Deploy
-      uses: peaceiris/actions-gh-pages@v3.6.1
+      uses: peaceiris/actions-gh-pages@v3.8.0
```
ocefpaf, line 50:
```suggestion
      uses: peaceiris/actions-gh-pages@v3
```

.github/workflows/publish.yml
```diff
@@ -6,10 +6,10 @@ jobs:
   packages:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3.1.0
```
ocefpaf, line 9:
```suggestion
    - uses: actions/checkout@v3.1.0
```
```suggestion
    - uses: actions/checkout@v3
```

.github/workflows/publish.yml
```diff
@@ -6,10 +6,10 @@ jobs:
...
 
     - name: Set up Python
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v4.3.0
```
ocefpaf, line 12:
```suggestion
      uses: actions/setup-python@v4
```

ocefpaf:
The error here is unrelated to this PR and we can investigate it later. Merging this for now.

.github/workflows/publish.yml
```diff
@@ -39,7 +39,7 @@ jobs:
 
     - name: Publish a Python distribution to PyPI
       if: ${{ github.event_name == 'release' }}
-      uses: pypa/gh-action-pypi-publish@master
+      uses: pypa/gh-action-pypi-publish@v1
```
ocefpaf, line 42: I believe we need to pin this one to the exact version unlike the others.

```suggestion
      uses: pypa/gh-action-pypi-publish@v1.5.1
```

PR #274, opened 2022-10-14
Fix segfaults

ocefpaf:
Hopefully this will fix the segfaults in our tests.


ocefpaf:
That seems to have fixed the segfaults! Yey! Next I'll work on the multiple server test that is being block by the server and then I'll move to @vinisalazar's PR. Sorry for the delay here Vini!

PR #276, opened 2022-10-27
Add Python 3.11 support

abkfenris:
Starts testing against Python 3.11 and adds a classifier for it.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,10 +24,10 @@ jobs:
 
     - name: Python ${{ matrix.python-version }}
       shell: bash -l {0}
-      run: |
+      run: >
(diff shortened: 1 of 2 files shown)


abkfenris:
Hmm, micromamba env setup is failing because it's [trying to use a Python 2.7 package](https://github.com/ioos/erddapy/actions/runs/3341325526/jobs/5532423985#step:4:22). It's not however failing appropriately on that step, which is probably a micromamba issue.

ocefpaf:
It is missing a mypy dependency on py3.11 micromamba is failing to fail in the env creation.

abkfenris:
Now its succeeding to fail on env creation.

ocefpaf:
mypy is available but now iris is the blocker.

PR #277, opened 2022-10-29
Interfaces

vinisalazar:
Hi,

this PR creates module `core.interfaces`, with methods to use URLs to generate third-party datasets. These methods are used in the `ERDDAP` class.

I rebased the `classes` branch from #267 with this branch.

If you could pay special attention to review how `kwargs` are being propagated across methods, that would be great.

Thanks,
Vini

### DIFF
--- a/erddapy/core/interfaces.py
+++ b/erddapy/core/interfaces.py
@@ -17,21 +17,19 @@ from erddapy.core.url import urlopen
 def to_pandas(url: str, requests_kwargs=None, **kw) -> pd.DataFrame:
     """Convert a URL to Pandas DataFrame."""
     if requests_kwargs is None:
-        requests_kwargs = dict()
+        requests_kwargs = {}
     data = urlopen(url, **requests_kwargs)
     try:
         return pd.read_csv(data, **kw)
     except Exception as e:
-        print("Couldn't process response into Pandas DataFrame.")
-        print(f"{type(e)} occurred. Please see below for the traceback.")
-        raise
+        raise ValueError(f"Could not read url {url} with Pandas.read_csv.") from e
 
 
 def to_ncCF(url: str, protocol: str = None, **kw) -> Dataset:
     """Convert a URL to a netCDF4 Dataset."""
     if protocol == "griddap":
         raise ValueError(
-            f"Cannot use ncCF with griddap. The URL you tried to access is: '{url}'.",
+            f"Cannot use .ncCF with griddap protocol. The URL you tried to access is: '{url}'.",
         )
     auth = kw.pop("auth", None)
     return _nc_dataset(url, auth=auth, **kw)
--- a/erddapy/core/interfaces.py
+++ b/erddapy/core/interfaces.py
@@ -30,7 +30,9 @@ def to_pandas(url: str, requests_kwargs=None, **kw) -> pd.DataFrame:
 def to_ncCF(url: str, protocol: str = None, **kw) -> Dataset:
     """Convert a URL to a netCDF4 Dataset."""
     if protocol == "griddap":
-        raise ValueError("Cannot use ncCF with griddap.")
+        raise ValueError(
+            f"Cannot use ncCF with griddap. The URL you tried to access is: '{url}'.",
+        )
     auth = kw.pop("auth", None)
     return _nc_dataset(url, auth=auth, **kw)
 
--- a/erddapy/core/interfaces.py
+++ b/erddapy/core/interfaces.py
@@ -38,7 +38,7 @@ def to_xarray(url: str, response="opendap", **kw) -> xr.Dataset:
         return xr.open_dataset(xr.backends.NetCDF4DataStore(nc), **kw)
 
 
-def to_iris(url: str, **kw) -> iris.CubeList:
+def to_iris(url: str, **kw):
     """Convert a URL to an iris CubeList."""
     data = urlopen(url, **kw)
     with _tempnc(data) as tmp:
--- a/erddapy/core/interfaces.py
+++ b/erddapy/core/interfaces.py
@@ -14,9 +14,9 @@ from erddapy.core.netcdf import _nc_dataset, _tempnc
 from erddapy.core.url import urlopen
 
 
-def to_pandas(url: str, **kw) -> pd.DataFrame:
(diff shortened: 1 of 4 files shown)


vinisalazar:
PS. please add the `hacktoberfest` labels :)

erddapy/erddapy.py
```diff
@@ -344,50 +344,37 @@ def to_pandas(self, **kw):
         """
         response = kw.pop("response", "csvp")
         url = self.get_download_url(response=response, **kw)
-        data = urlopen(url, auth=self.auth, **self.requests_kwargs)
-        return pd.read_csv(data, **kw)
+        return to_pandas(url, **kw)
 
     def to_ncCF(self, **kw):
         """Load the data request into a Climate and Forecast compliant netCDF4-python object."""
         if self.protocol == "griddap":
```
ocefpaf, line 351: This check should probably be in the function you created and not here in the method, right?

ocefpaf: Maybe you can perform that check on the URL to avoid pushing the url creation to the function.

vinisalazar: I was thinking about how to do this but the check is required to determine the `response` argument in the URL creation method.

ocefpaf:
> PS. please add the `hacktoberfest` labels :)

Sorry! Missed that completely and we are past the deadline!! Please ping me directly next time.

erddapy/erddapy.py
ocefpaf: The one in `interfaces.py` would be broken if someone tries to use it directly .I believe that, to avoid this, we are OK adding an extra keyword `protocol=None` here that is passed there. Then we need a check here like:

```python
protocol = protocol if protocol else self.protocol
```

erddapy/core/interfaces.py
```diff
@@ -0,0 +1,54 @@
...
+        nc = _nc_dataset(url, auth=auth, **kw)
+        return xr.open_dataset(xr.backends.NetCDF4DataStore(nc), **kw)
+
+
+def to_iris(url: str, **kw):
+    """Convert a URL to an iris CubeList."""
+    data = urlopen(url, **kw)
+    with _tempnc(data) as tmp:
+        cubes = iris.load_raw(tmp, **kw)
+        _ = [cube.data for cube in cubes]
```
vinisalazar, line 53: @ocefpaf Iris version is pinned in the `environment.yml` file, does it need to be anywhere else?

erddapy/core/interfaces.py
```diff
@@ -0,0 +1,54 @@
...
+Interface between URL responses and third-party libraries.
+
+This module takes an URL or the bytes response of a request and converts it to Pandas,
+XArray, Iris, etc. objects.
+"""
+
+import iris
+import pandas as pd
+import xarray as xr
+from netCDF4 import Dataset
```
vinisalazar, line 11: Sorry, I wasn't sure if we had settled on using `Dataset` or `import netCDF4 as nc` and then use `nc.Dataset` (like `xr.Dataset`).

vinisalazar: I would prefer to avoid importing the whole module for the sake of a single type hint (even if it improves readability).

vinisalazar (commented):
Hi @ocefpaf @abkfenris, I addressed all comments that we had noted down, so I think this is ready to be reviewed again.

erddapy/core/interfaces.py
ocefpaf: This is fine. We don't have any clashes with `Dataset` in this module.

erddapy/core/interfaces.py
```diff
@@ -8,26 +8,29 @@
...
 
 from erddapy.core.netcdf import _nc_dataset, _tempnc
 from erddapy.core.url import urlopen
 
 
-def to_pandas(url: str, requests_kwargs=dict(), **kw) -> pd.DataFrame:
+def to_pandas(url: str, requests_kwargs=None, **kw) -> pd.DataFrame:
     """Convert a URL to Pandas DataFrame."""
+    if requests_kwargs is None:
+        requests_kwargs = dict()
```
ocefpaf, line 20: It is kind of standard to initialize dicts like this. I know, less readable, but there are some optimizations at play here.

```suggestion
        requests_kwargs = {}
```

erddapy/core/interfaces.py
```diff
@@ -8,26 +8,29 @@
...
     data = urlopen(url, **requests_kwargs)
     try:
         return pd.read_csv(data, **kw)
-    except Exception:
+    except Exception as e:
         print("Couldn't process response into Pandas DataFrame.")
+        print(f"{type(e)} occurred. Please see below for the traceback.")
```
ocefpaf, line 26: Ideally we should narrow down the scope here instead of casting a wide net with `Exception`. What errors are we expecting?

In a way, we could interpret it as a ValueError to the `pandas_csv` and do a raise from like:

```python
try:
    return pd.read_csv(data, **kw)
except Exception as e:
    raise ValueError(f"Could not read url {url} with Pandas.read_csv.") from e
```

erddapy/core/interfaces.py
```diff
@@ -8,26 +8,29 @@
...
         raise
 
 
-def to_ncCF(url: str, protocol: str = None, **kw) -> ncDataset:
+def to_ncCF(url: str, protocol: str = None, **kw) -> Dataset:
     """Convert a URL to a netCDF4 Dataset."""
     if protocol == "griddap":
-        return ValueError("Cannot use ncCF with griddap.")
+        raise ValueError("Cannot use ncCF with griddap.")
```
ocefpaf:
```suggestion
        raise ValueError("Cannot use .ncCF with griddap protocol.")
```

erddapy/core/interfaces.py
ocefpaf: We need to change it in https://github.com/ioos/erddapy/blob/3119ee22ffd7d66d6b92eb28c5feccc20a4d91d3/requirements-dev.txt#L31.

erddapy/core/interfaces.py
vinisalazar: Thanks, that looks much better

ocefpaf (approved):
LGTM. @abkfenris any final comments?

abkfenris (approved):
Looks good to me

PR #280, opened 2022-11-19
Fix kwargs propagation in interfaces module

vinisalazar:
Hi,

This is something I caught while running the `ds = e.to_xarray(decode_times=False)` line in `02-extras.ipynb`.

**Summary of changes**
  - Fix kwargs propagation for urlopen and third-party libraries
  - Improve docstrings
  - Add typing.Dict for type annotation

### DIFF
--- a/erddapy/core/interfaces.py
+++ b/erddapy/core/interfaces.py
@@ -63,8 +63,7 @@ def to_ncCF(
         )
     if requests_kwargs is None:
         requests_kwargs = {}
-    auth = requests_kwargs.pop("auth", None)
-    return _nc_dataset(url, auth=auth, **requests_kwargs)
+    return _nc_dataset(url, **requests_kwargs)
 
 
 def to_xarray(
@@ -78,18 +77,17 @@ def to_xarray(
 
     url: URL to request data from.
     response: type of response to be requested from the server.
-    requests_kwargs: arguments to be passed to urlopen method.
+    requests_kwargs: arguments to be passed to urlopen method (including auth)
     xarray_kwargs: kwargs to be passed to third-party library (xarray).
     """
     import xarray as xr
 
     if requests_kwargs is None:
         requests_kwargs = {}
-    auth = requests_kwargs.pop("auth", None)
     if response == "opendap":
         return xr.open_dataset(url, **xarray_kwargs)
     else:
-        nc = _nc_dataset(url, auth=auth, **requests_kwargs)
+        nc = _nc_dataset(url, **requests_kwargs)
         return xr.open_dataset(xr.backends.NetCDF4DataStore(nc), **xarray_kwargs)
 
 
--- a/erddapy/core/interfaces.py
+++ b/erddapy/core/interfaces.py
@@ -6,6 +6,8 @@ XArray, Iris, etc. objects.
 """
 from typing import TYPE_CHECKING
 
+from typing import Dict
+
 import pandas as pd
 
 from erddapy.core.netcdf import _nc_dataset, _tempnc
@@ -15,9 +17,14 @@ if TYPE_CHECKING:
     import xarray as xr
     from netCDF4 import Dataset
 
+def to_pandas(url: str, requests_kwargs: Dict = None, **kw) -> "pd.DataFrame":
+    """
+    Convert a URL to Pandas DataFrame.
 
-def to_pandas(url: str, requests_kwargs=None, **kw) -> "pd.DataFrame":
-    """Convert a URL to Pandas DataFrame."""
+    url: URL to request data from.
+    requests_kwargs: arguments to be passed to urlopen method.
+    **kw: kwargs to be passed to third-party library (pandas).
+    """
     if requests_kwargs is None:
         requests_kwargs = {}
     data = urlopen(url, **requests_kwargs)
@@ -37,23 +44,42 @@ def to_ncCF(url: str, protocol: str = None, **kw) -> "Dataset":
     return _nc_dataset(url, auth=auth, **kw)
 
 
-def to_xarray(url: str, response="opendap", **kw) -> "xr.Dataset":
-    """Convert a URL to an xarray dataset."""
+def to_xarray(
+    url: str, response="opendap", requests_kwargs: Dict = None, **kw
+) -> "xr.Dataset":
+    """
+    Convert a URL to an xarray dataset.
+
+    url: URL to request data from.
+    response: type of response to be requested from the server.
+    requests_kwargs: arguments to be passed to urlopen method.
+    **kw: kwargs to be passed to third-party library (xarray).
+    """
     import xarray as xr
 
     auth = kw.pop("auth", None)
     if response == "opendap":
         return xr.open_dataset(url, **kw)
     else:
-        nc = _nc_dataset(url, auth=auth, **kw)
+        if requests_kwargs is None:
+            requests_kwargs = {}
+        nc = _nc_dataset(url, auth=auth, **requests_kwargs)
         return xr.open_dataset(xr.backends.NetCDF4DataStore(nc), **kw)
 
 
-def to_iris(url: str, **kw):
-    """Convert a URL to an iris CubeList."""
+def to_iris(url: str, requests_kwargs: Dict = None, **kw):
+    """
+    Convert a URL to an iris CubeList.
+
+    url: URL to request data from.
+    requests_kwargs: arguments to be passed to urlopen method.
+    **kw: kwargs to be passed to third-party library (iris).
+    """
     import iris
 
-    data = urlopen(url, **kw)
+    if requests_kwargs is None:
+        requests_kwargs = {}
+    data = urlopen(url, **requests_kwargs)
     with _tempnc(data) as tmp:
         cubes = iris.load_raw(tmp, **kw)
         _ = [cube.data for cube in cubes]
--- a/erddapy/core/interfaces.py
+++ b/erddapy/core/interfaces.py
@@ -17,35 +17,61 @@ if TYPE_CHECKING:
     import xarray as xr
     from netCDF4 import Dataset
 
+<<<<<<< HEAD
 def to_pandas(url: str, requests_kwargs: Optional[Dict] = None, **kw) -> "pd.DataFrame":
+=======
+def to_pandas(
+    url: str, requests_kwargs: Optional[Dict] = None, **pandas_kwargs
+) -> pd.DataFrame:
+>>>>>>> bf7cbb2 (Renaming **kw according to each library)
     """
     Convert a URL to Pandas DataFrame.
 
     url: URL to request data from.
     requests_kwargs: arguments to be passed to urlopen method.
-    **kw: kwargs to be passed to third-party library (pandas).
+    **pandas_kwargs: kwargs to be passed to third-party library (pandas).
     """
     if requests_kwargs is None:
         requests_kwargs = {}
     data = urlopen(url, **requests_kwargs)
     try:
-        return pd.read_csv(data, **kw)
+        return pd.read_csv(data, **pandas_kwargs)
     except Exception as e:
         raise ValueError(f"Could not read url {url} with Pandas.read_csv.") from e
 
 
+<<<<<<< HEAD
 def to_ncCF(url: str, protocol: str = None, **kw) -> "Dataset":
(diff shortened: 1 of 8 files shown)


erddapy/core/interfaces.py
```diff
@@ -14,8 +16,14 @@
 from erddapy.core.url import urlopen
 
 
-def to_pandas(url: str, requests_kwargs=None, **kw) -> pd.DataFrame:
-    """Convert a URL to Pandas DataFrame."""
+def to_pandas(url: str, requests_kwargs: Dict = None, **kw) -> pd.DataFrame:
```
abkfenris, line 19: I think any time `:Dict = None` is in function args/kwargs, it needs to be specified as an optional, so `request_kwargs: Optional[Dict] = None` with `Optional` also imported from `typing`.

Otherwise mypy will get cranky on folks downstream when they don't want to specify `request_kwargs`. 

Didn't we also discuss breaking out pandas kwargs explicitly, rather than collecting whatever is left?

vinisalazar: Just added the `Optional[Dict]` annotations.

I kept the Pandas kwargs as `**kw` for sake of consistency with the other methods, should I change all `**kw` to `pandas_kwargs`, `xarray_kwargs`, and so on?

ocefpaf: Agreed. More declarative and easier to follow.

erddapy/core/interfaces.py
```diff
@@ -26,7 +34,14 @@ def to_pandas(url: str, requests_kwargs=None, **kw) -> pd.DataFrame:
...
 
 def to_ncCF(url: str, protocol: str = None, **kw) -> Dataset:
-    """Convert a URL to a netCDF4 Dataset."""
+    """
+    Convert a URL to a netCDF4 Dataset.
+
+    url: URL to request data from.
+    protocol: 'griddap' or 'tabledap'.
+    requests_kwargs: arguments to be passed to urlopen method.
+    **kw: kwargs to be passed to third-party library (ncCF).
```
ocefpaf, line 43: These are requests kwargs, right?

vinisalazar: Yes, that's right. I'll fix that docstring.

erddapy/core/interfaces.py
```diff
@@ -14,41 +16,83 @@
...
+    """
     if protocol == "griddap":
         raise ValueError(
             f"Cannot use .ncCF with griddap protocol. The URL you tried to access is: '{url}'.",
         )
-    auth = kw.pop("auth", None)
-    return _nc_dataset(url, auth=auth, **kw)
+    if requests_kwargs is None:
+        requests_kwargs = {}
+    auth = requests_kwargs.pop("auth", None)
```
vinisalazar, line 56: Should we remove this in favour of just passing auth as a key in the `request_kwargs` dict?

erddapy/core/interfaces.py
```diff
@@ -14,41 +16,83 @@
...
-    """Convert a URL to an xarray dataset."""
-    auth = kw.pop("auth", None)
+    url: URL to request data from.
+    response: type of response to be requested from the server.
+    requests_kwargs: arguments to be passed to urlopen method.
+    xarray_kwargs: kwargs to be passed to third-party library (xarray).
+    """
+    if requests_kwargs is None:
+        requests_kwargs = {}
+    auth = requests_kwargs.pop("auth", None)
```
vinisalazar, line 76: Same as in L56, should we remove this in favour of getting auth as a key in the `requests_kwargs` dict?

erddapy/core/interfaces.py
ocefpaf: Yeah. Before we had both as a single `kw` and needed this workaround. We can do better now!

erddapy/core/interfaces.py
ocefpaf: Yep.

erddapy/core/interfaces.py
```diff
@@ -68,16 +67,15 @@ def to_xarray(
 
     url: URL to request data from.
     response: type of response to be requested from the server.
-    requests_kwargs: arguments to be passed to urlopen method.
+    requests_kwargs: arguments to be passed to urlopen method (including auth)
```
ocefpaf, line 70: We probably don't need this. We can add an example that has the `auth` dict inside the `requests_kwargs` dict to illustrate that though. What do you think?

```suggestion
    requests_kwargs: arguments to be passed to urlopen method
```

vinisalazar: I agree. I will add that example when I work on the docs again.

erddapy/core/interfaces.py
```diff
@@ -28,6 +30,8 @@ def to_pandas(
     """
     if requests_kwargs is None:
         requests_kwargs = {}
+    if pandas_kwargs is None:
+        pandas_kwargs = {}
     data = urlopen(url, **requests_kwargs)
```
ocefpaf, line 35: Let's drop the `**` everywhere, also in the docs.

vinisalazar: Ah, ok, I misunderstood.

ocefpaf: No problem. I'm slow to see all the changes required here too. Reviewing mostly from a tiny cellphone screen these days :-/

erddapy/core/interfaces.py
```diff
@@ -14,41 +16,93 @@
...
+    url: URL to request data from.
+    requests_kwargs: arguments to be passed to urlopen method.
+    **pandas_kwargs: kwargs to be passed to third-party library (pandas).
+    """
     if requests_kwargs is None:
         requests_kwargs = {}
-    data = urlopen(url, **requests_kwargs)
+    if pandas_kwargs is None:
+        pandas_kwargs = {}
+    data = urlopen(url, requests_kwargs)
```
ocefpaf, line 35: We don't need to initialize an empty dict for requests_kwargs, right? That one is optional in `urlopen ` .

erddapy/core/interfaces.py
```diff
@@ -14,41 +16,93 @@
...
+    """
     if protocol == "griddap":
         raise ValueError(
             f"Cannot use .ncCF with griddap protocol. The URL you tried to access is: '{url}'.",
         )
-    auth = kw.pop("auth", None)
-    return _nc_dataset(url, auth=auth, **kw)
+    if requests_kwargs is None:
+        requests_kwargs = {}
+    return _nc_dataset(url, requests_kwargs)
```
ocefpaf, line 60: Same comment from https://github.com/ioos/erddapy/pull/280/files#r1035949714.

erddapy/core/interfaces.py
```diff
@@ -14,41 +16,93 @@
...
-def to_xarray(url: str, response="opendap", **kw) -> xr.Dataset:
-    """Convert a URL to an xarray dataset."""
-    auth = kw.pop("auth", None)
+    url: URL to request data from.
+    response: type of response to be requested from the server.
+    requests_kwargs: arguments to be passed to urlopen method.
+    xarray_kwargs: kwargs to be passed to third-party library (xarray).
+    """
+    if requests_kwargs is None:
+        requests_kwargs = {}
```
ocefpaf, line 78: We probably do not need these lines.

vinisalazar:
Thanks for the code review, please have a look at https://github.com/ioos/erddapy/pull/280/commits/7c49ef654d99ff95fa844ef9ca70b44884fb118d too, I was having this problem before and I think this may be a good fix?

ocefpaf:
@vinisalazar and @abkfenris I rebased this one and, hopefully, I did not screw it up! Do you mind taking a second look so we can merge this one?

abkfenris:
Hmm, I think the doc test failures we're seeing are the difference in call signatures between the various `to_X()` functions and methods. 

https://github.com/ioos/erddapy/blob/f11ad169713e1fa470a51c642d06411779e99f61/erddapy/erddapy.py#L334-L347

vs

https://github.com/vinisalazar/erddapy/blob/0ee3c962996dc911d1a0a39c85bc6318ca4b5eef/erddapy/erddapy.py#L334-L353

What is more worrying is that they aren't being caught by other tests. While the new method matches up better to the underlying functions it breaks things for folks currently using `erddapy.ERDDAP`. Either we need to accept a breaking change, or do some smart shuffling of `**kwargs` around between methods and functions.

vinisalazar:
Thanks @abkfenris for pointing that out.

I found one use case in [gliderpy](https://github.com/ioos/gliderpy/blob/bb5703d9c91729247831f6e9ef2c439e3da4c4c3/gliderpy/fetchers.py#L66) which would surely break.

Could this be a possible fix? It preserves the old `ERDDAP.to_pandas` method while supporting the new `interfaces.to_pandas` kwargs propagation:

```python
    def to_pandas(
        self,
        **kw
    ):
        """Save a data request to a pandas.DataFrame.

        Accepts any `pandas.read_csv` keyword arguments, passed as a dictionary to pandas_kwargs.

        This method uses the .csvp [1] response as the default for simplicity,
        please check ERDDAP's documentation for the other csv options available.

        [1] Download a ISO-8859-1 .csv file with line 1: name (units). Times are ISO 8601 strings.

        """
        url = self.get_download_url(response=response, **kw)
        return to_pandas(url, pandas_kwargs=dict(**kw))
```

vinisalazar:
I had to do some updates to the documentation so it would build correctly.

vinisalazar:
Not sure what to do about the current error, since it's a problem of being black listed from the ERDDAP server.

I could save the data on a file and add that file to version control (it's small at 44K), and we load the data from the file to build the documentation, and leave a comment for users running the notebook to actually make the request.

vinisalazar:
e.g.

```python
import xarray as xr

ds = xr.open_dataset("jplAvisoSshMon_SA.nc")

# We have this data saved to spare the server from the request
# You can actually request the dataset using the following line: 
# ds = e.to_xarray()
```

abkfenris (commented):
I wonder if we don't have enough tests on the existing `erddapy.ERDDAP` class to make it clear when we break backwards compatibility. 

My first thought was that we probably changed a test since the last release and that why tests didn't catch things that the docs did, but it doesn't seem to really have any major differences. https://github.com/vinisalazar/erddapy/compare/v0.9.0...fix-interfaces-kwargs

erddapy/erddapy.py
```diff
@@ -346,17 +342,15 @@ def to_pandas(
...
 
         """
-        if requests_kwargs is None:
-            requests_kwargs = {}
-        response = requests_kwargs.pop("response", "csvp")
-        url = self.get_download_url(response=response, **requests_kwargs)
-        return to_pandas(url, requests_kwargs, pandas_kwargs)
+        response = kw.pop("response", "csvp")
+        url = self.get_download_url(response=response, **kw)
+        return to_pandas(url, pandas_kwargs=dict(**kw))
```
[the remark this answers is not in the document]
vinisalazar: That's understandable, I just kept it how it is in the `main` branch at the moment.

vinisalazar:
> My first thought was that we probably changed a test since the last release and that why tests didn't catch things that the docs did, but it doesn't seem to really have any major differences.

The difference is that now the CI tries to build the docs, where it didn't before (please correct me if I'm wrong @ocefpaf).

abkfenris:
> > My first thought was that we probably changed a test since the last release and that why tests didn't catch things that the docs did, but it doesn't seem to really have any major differences.
> 
> The difference is that now the CI tries to build the docs, where it didn't before (please correct me if I'm wrong @ocefpaf).

Sorry, I was trying to say, our regular tests have not been exercising the API the right way if we are only getting errors with the docs build. We should have caught that with pytest.

vinisalazar:
Currently tests are failing because of the blocked IP address, I believe that the docs would build correctly if that was done locally. Once we fix this and merge this PR, I can submit another one addressing tests and test coverage.

vinisalazar:
Is there anything else I could do so we can get this one merged?

vinisalazar:
I believe this should pass after #295 is merged

vinisalazar:
@ocefpaf we are getting a `ReadTimeout` error when building the docs... any clue on why this is happening/what could be done?

ocefpaf:
I'll try to fix the docs later today. Let's get this one in! Sorry for the delay Vini!

PR #281, opened 2022-11-19
Add public methods to erddapy module

vinisalazar:
Hi,

while working on the docs for the core interface, I realised that the URL parsing and interfaces methods would have to be directly imported from their respective modules, rather than being accessed from the `erddapy` module. This seemed a bit unintuitive, so I added public methods directly to the `erddapy.__init__` module.

Users can then run:
```python
from erddapy import get_download_url, to_pandas

url = get_download_url(...)
df = to_pandas(url)

# or
import erddapy

url = erddapy.get_download_url(...)
df = erddapy.to_pandas(url)
```

instead of

```python
from erddapy.core.url import get_download_url
from erddapy.core.interfaces import to_pandas
```

which seems much harder to do.

**Summary of changes:**
  - Add public methods to `erddapy.__init__`

  This allows using the URL parsing and interfaces methods directly from the erddapy module,
  without having to access the core.url and core.interfaces modules.

### DIFF
--- a/erddapy/__init__.py
+++ b/erddapy/__init__.py
@@ -1,9 +1,31 @@
 """Easier access to scientific data."""
 
+from erddapy.core.interfaces import to_iris, to_ncCF, to_pandas, to_xarray
+from erddapy.core.url import (
+    get_categorize_url,
+    get_download_url,
+    get_info_url,
+    get_search_url,
+    parse_dates,
+    urlopen,
+)
 from erddapy.erddapy import ERDDAP
 from erddapy.servers.servers import servers
 
-__all__ = ["ERDDAP", "servers"]
+__all__ = [
+    "ERDDAP",
+    "servers",
+    "get_categorize_url",
+    "get_download_url",
+    "get_info_url",
+    "get_search_url",
+    "parse_dates",
+    "urlopen",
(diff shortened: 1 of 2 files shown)


abkfenris:
If we are going to expose them all at one spot, I'd rather expose them at `erddapy.core` rather than at the top level. 

I think it would help clarify to users that refactoring is going on, and allow them to make an explicit choice which world to live in. It also would decrease the amount that is in a single namespace.

```py
import erddapy.core

url = erddapy.core.get_download_url(...)
```

ocefpaf:
Yep. I agreed with @abkfenris, putting them all at the root level would be a bit confusing for folks browsing the module. Having them into a `.core` is probably the best place for them.

vinisalazar:
Done with https://github.com/ioos/erddapy/pull/281/commits/6fdb280abfc8dccce9f1f1c043723add642b33e6.

Users can run:

```python
import erddapy

erddapy.core.get_download_url()
erddapy.servers
...
```

i.e. no need to `import erddapy.core`.

ocefpaf (approved):
I'll wait until Friday just in case Alex see this before then.

abkfenris (approved):

PR #292, opened 2022-12-07
Fix optional deps

### LINKED ISSUE #291: Refactor tests and mandatory vs optional dependencies
The refactor ended up making netCDF4, iris, and xarray mandatory dependencies. However, I'd like to keep the core of erddapy as simple as possible (httpx and pandas only, dropping even pytz when we move to Python >=3.9).

- [x] add a "core" set of tests that do not require the optional dependencies
- [x] add a "full" set of tests that with all the dependencies
- [x] use https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING  to avoid type checking when the optional packages are not available
- [x] move type checks that require 3rd party import to string literals
- [ ] drop py37 and py38 support

ocefpaf:
Closes #291 and incorporates #276.


.github/workflows/tests_core.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: "3.x"
+
+    - name: Core Tests
+      run: >
+        python -m pip install .
+        && python -m pip install pytest
+        && python -m pytest -rxs tests/test_erddapy.py
```
abkfenris, line 27: Maybe split out the install into a separate step? Makes it a bit easier to figure out which step failed.

Should these also be tested against a matrix of versions?

```suggestion

    - name: Install
      run: >
        python -m pip install .
        && python -m pip install pytest
    
    - name: Core Tests
      run:  python -m pytest -rxs tests/test_erddapy.py
```

abkfenris (approved):
I think it looks ok, but it may be easier to tell when we do have failures if the steps in the core test are broken apart. 

It may also be worth combining the core tests in as a new job in the current `tests.yml`, as it can be a bit quicker to navigate between different jobs in the same workflow than between jobs in different workflows.

.github/workflows/tests_core.yml
ocefpaf: > Should these also be tested against a matrix of versions?

It is not needed IMO b/c we do repeat these tests in the full test matrix. The goal of this one is to catch issues in a non-conda env (like the missing httpx! :fearful:) and avoid regressions with the optional imports.

ocefpaf:
> It may also be worth combining the core tests in as a new job in the current `tests.yml`, as it can be a bit quicker to navigate between different jobs in the same workflow than between jobs in different workflows.

I don't have a strong opinion on this but I'm slightly more inclined to multiple files than a huge one. Sure, it won't be a huge one right away but they can grow in size. In this cases the context is not only a small subset without the optional deps but also a non-conda setup.

ocefpaf:
I'll work on the docs in another PR. Let's get this one in.

PR #295, opened 2023-02-17
Re-write docs to avoid bad servers

ocefpaf:
Change the servers we hit when building the docs to stable ones. Or at least ones that won't kick us out for just building docs :-/


ocefpaf:
We have docs again :tada:

PR #298, opened 2023-03-31
Use more stable servers

ocefpaf:
I simplified the quick intro a lot to fix a more stable server. Fingers crossed!


ocefpaf:
It looks like the change in the caching documented in https://github.com/ioos/erddapy/pull/297#issuecomment-1491920905 did break something.

```
"""
Traceback (most recent call last):
  File "/home/runner/micromamba-root/envs/TEST/lib/python3.11/site-packages/joblib/externals/loky/backend/queues.py", line 125, in _feed
    obj_ = dumps(obj, reducers=reducers)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/micromamba-root/envs/TEST/lib/python3.11/site-packages/joblib/externals/loky/backend/reduction.py", line 211, in dumps
    dump(obj, buf, reducers=reducers, protocol=protocol)
  File "/home/runner/micromamba-root/envs/TEST/lib/python3.11/site-packages/joblib/externals/loky/backend/reduction.py", line 204, in dump
    _LokyPickler(file, reducers=reducers, protocol=protocol).dump(obj)
  File "/home/runner/micromamba-root/envs/TEST/lib/python3.11/site-packages/joblib/externals/cloudpickle/cloudpickle_fast.py", line 632, in dump
    return Pickler.dump(self, obj)
           ^^^^^^^^^^^^^^^^^^^^^^^
_pickle.PicklingError: Can't pickle <functools._lru_cache_wrapper object at 0x7fd6153a0bf0>: it's not the same object as erddapy.erddapy.ERDDAP._get_variables_uncached
"""
```

However, the question is if we should be accessing a server like that or not. It has been discourage by data providers, so maybe we should remove that kind of parallel access from our docs altogether.

Issue #128, opened 2020-07-30
ImportError: cannot import name 'parse_time_string' from 'pandas.core.tools.datetimes'


emiliom:
I just built a conda environment today with erddapy and pandas, among other things. No version pinning at all. Python 3.8.5, pandas 1.1.0, erddapy 0.5.3. Just importing erddapy produces the pandas error below.

Same issue when I built the conda env while pinning Python to 3.7.

Any suggestions for pinning a specific version of Pandas, erddapy or both to get past this issue? Thanks!

```python
~/miniconda/envs/nvsgliderapp/lib/python3.8/site-packages/erddapy/__init__.py in <module>
----> 1 from erddapy.erddapy import ERDDAP, servers
      2 
      3 __all__ = ["ERDDAP", "servers"]
      4 
      5 from ._version import get_versions

~/miniconda/envs/nvsgliderapp/lib/python3.8/site-packages/erddapy/erddapy.py in <module>
     13 import pandas as pd
     14 
---> 15 from erddapy.utilities import (
     16     _check_url_response,
     17     _tempnc,

~/miniconda/envs/nvsgliderapp/lib/python3.8/site-packages/erddapy/utilities.py in <module>
     11 import pytz
     12 import requests
---> 13 from pandas.core.tools.datetimes import parse_time_string
     14 
     15 _server = namedtuple("server", ["description", "url"])

ImportError: cannot import name 'parse_time_string' from 'pandas.core.tools.datetimes' (/home/mayorga/miniconda/envs/nvsgliderapp/lib/python3.8/site-packages/pandas/core/tools/datetimes.py)
```

emiliom:
FYI, I tried to find on what version `pandas.core.tools.datetimes.parse_time_string` was deprecated, but couldn't find it. What I was able to do is confirm that it's present in pandas 1.0.1, so I'm going with that (pinning my conda env to that pandas version).

This statement from the pandas API docs is worth noting:

**Warning**
The `pandas.core`, `pandas.compat`, and `pandas.util` top-level modules are PRIVATE. Stable functionality in such modules is not guaranteed.

Issue #34, opened 2018-04-20
ImportError: No module named tslibs.parsing


BrianOBlanton:
Hi,  when importing erddapy, I get this: 

**from erddapy import ERDDAP**
...
...
**ImportError: No module named tslibs.parsing**

Is this a pandas problem?    Here's the complete traceback: 

```python 
ImportErrorTraceback (most recent call last)
<ipython-input-3-20c6c8f35979> in <module>()
----> 1 from erddapy import ERDDAP

/home/bblanton/.local/lib/python2.7/site-packages/erddapy/__init__.py in <module>()
      1 from __future__ import (absolute_import, division, print_function)
      2 
----> 3 from erddapy.erddapy import ERDDAP, servers
      4 
      5 __all__ = [

/home/bblanton/.local/lib/python2.7/site-packages/erddapy/erddapy.py in <module>()
      8 from __future__ import (absolute_import, division, print_function)
      9 
---> 10 from erddapy.url_builder import (
     11     download_url,
     12     info_url,

/home/bblanton/.local/lib/python2.7/site-packages/erddapy/url_builder.py in <module>()
     18     from urllib import quote_plus
     19 
---> 20 from erddapy.utilities import (_check_url_response, parse_dates, quote_string_constraints)
     21 
     22 

/home/bblanton/.local/lib/python2.7/site-packages/erddapy/utilities.py in <module>()
     10 from collections import namedtuple
     11 
---> 12 from pandas._libs.tslibs.parsing import parse_time_string
     13 
     14 import pytz

ImportError: No module named tslibs.parsing
```

ocefpaf:
Can you check the version of pandas that you have installed there?
You can copy and paste this in the command line:

```shell
python -c 'import pandas; print(pandas.__version__)'
```

BrianOBlanton:
0.20.3

ocefpaf:
Can you update to `0.22.0`? I'll try to add some backwards compatibility, I never realized that `parse_time_string` was so new.

BrianOBlanton:
Yes, I will., and let you know.

BrianOBlanton:
that worked! thanks.  I'll start getting data...

ocefpaf:
Awesome! I'll leave this open to remind me to add support for older pandas.

BrianOBlanton:
OK. Thanks again.


Cheers,

[...]

PR #300, opened 2023-04-04
pandas 2.0 support

### LINKED ISSUE #299: Impossible to import erddapy with pandas 2.0
I was building a conda environment with unpinned packages:
  - aiohttp
  - erddapy
  - fsspec
  - netCDF4
  - packaging
  - scipy
  - toolz
  - xarray
  - pandas
  - numpy

When conda/mamba solve this, it peaks the Pandas version 2.0
from which erddapy is trying to import ``parse_time_string`` from ``pandas._libs.tslibs.parsing``
but the ``parse_time_string`` function is not longer available there (apparently it was removed after [v2.0.0.dev0](https://github.com/pandas-dev/pandas/blob/v2.0.0.dev0/pandas/tests/tslibs/test_parsing.py))

following the function signature, it looks to have been replaced by: [parse_datetime_string_with_reso](https://github.com/pandas-dev/pandas/blob/v2.0.0/pandas/_libs/tslibs/parsing.pyi#L14), which was in fact called by ``parse_time_string`` code.

May be erddapy could simply rely on [py_parse_datetime_string](https://github.com/pandas-dev/pandas/blob/v2.0.0/pandas/_libs/tslibs/parsing.pyi#L9) in [here](https://github.com/ioos/erddapy/blob/9779b3f6bd9f8a60a0b271c49dc483b65f69f48b/erddapy/core/url.py#L128) ?

this looks like a recurrent issue for erddapy (#128 #34 #240 )

> At some point we should probably vendor parse_time_string b/c that is not part of pandas' public API.

ocefpaf:
Closes #299 

In this PR:
- We are still relying on a private function but that code is the best there is :-/
- We are pinning to pandas >=2.0. Should be OK b/c we are patching the previous released erddapies to work with older pandas only (conda-forge), there is nothing we can do for PyPI;
- Re-added the tag regex that was accidentally removed;
- Added the `dayfirst` and `yearfirst` options to `parse_dates`. Only the default work when calling it internally but a user has the option to convert the date themselves this way.

### DIFF
--- a/erddapy/core/url.py
+++ b/erddapy/core/url.py
@@ -9,9 +9,7 @@ from urllib.parse import quote_plus
 
 import httpx
 import pytz
-from pandas._libs.tslibs.parsing import (
-    py_parse_datetime_string as _py_parse_datetime_string,
-)
+from pandas import to_datetime
 
 ListLike = Union[List[str], Tuple[str]]
 OptionalStr = Optional[str]
@@ -129,11 +127,11 @@ def parse_dates(
 
     """
     if isinstance(date_time, str):
-        parse_date_time = _py_parse_datetime_string(
+        parse_date_time = to_datetime(
             date_time,
             dayfirst=dayfirst,
             yearfirst=yearfirst,
-        )
+        ).to_pydatetime()
     else:
         parse_date_time = date_time
 
--- a/erddapy/core/url.py
+++ b/erddapy/core/url.py
@@ -9,7 +9,9 @@ from urllib.parse import quote_plus
 
 import httpx
 import pytz
-from pandas._libs.tslibs.parsing import parse_time_string
+from pandas._libs.tslibs.parsing import (
+    py_parse_datetime_string as _py_parse_datetime_string,
+)
 
 ListLike = Union[List[str], Tuple[str]]
 OptionalStr = Optional[str]
@@ -113,7 +115,11 @@ def _check_substrings(constraint):
     return any([True for substring in substrings if substring in str(constraint)])
 
 
-def parse_dates(date_time: Union[datetime, str]) -> float:
+def parse_dates(
+    date_time: Union[datetime, str],
+    dayfirst=False,
+    yearfirst=False,
+) -> float:
     """
     Parse dates to ERDDAP internal format.
 
@@ -123,9 +129,11 @@ def parse_dates(date_time: Union[datetime, str]) -> float:
 
     """
     if isinstance(date_time, str):
-        # pandas returns a tuple with datetime, dateutil, and string representation.
-        # we want only the datetime obj.
-        parse_date_time = parse_time_string(date_time)[0]
+        parse_date_time = _py_parse_datetime_string(
+            date_time,
+            dayfirst=dayfirst,
+            yearfirst=yearfirst,
+        )
     else:
         parse_date_time = date_time
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,6 +5,7 @@ geopandas
 interrogate
 joblib
 jupyter
+nbconvert<7.3.0
 nbsphinx
 netcdf4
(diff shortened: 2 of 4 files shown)


erddapy/core/url.py
```diff
@@ -9,7 +9,7 @@
 
 import httpx
 import pytz
-from pandas._libs.tslibs.parsing import parse_time_string
+from pandas import to_datetime
```
ocefpaf, line 12: @gmaze I should've done this a long time ago :-)

Thanks for the ping in

> this looks like a recurrent issue for erddapy (https://github.com/ioos/erddapy/issues/128 https://github.com/ioos/erddapy/issues/34 https://github.com/ioos/erddapy/pull/240 )

requirements-dev.txt
```diff
@@ -5,6 +5,7 @@ geopandas
 interrogate
 joblib
 jupyter
+nbconvert<7.3.0
```
ocefpaf, line 8: See https://github.com/jupyter/nbconvert/issues/1970

requirements.txt
```diff
@@ -1,3 +1,3 @@
 httpx
-pandas>=0.20.3
+pandas>=1.5.3
```
ocefpaf, line 2: We could probably go lower but I don't think we need to.

erddapy/core/url.py
gmaze: awesome, thanks for the quick fix !

abkfenris: What about extracting the import into a function and then tweaking execution on import errors? 

Then we could probably revert the new pin.

```py
def dang_datetimes(date_time: Union[datetime, str], dayfirst: bool=False, yearfirst: bool=False):
    try:
        from pandas import to_datetime

        return to_datetime(datetime, dayfirst=dayfirst, yearfirst=yearfalse).to_pydatetime()
    except ImportError:
        from pandas._libs.tslibs.parsing import parse_time_string

        return parse_time_string(date_time)[0]
```

ocefpaf: > Then we could probably revert the new pin.

The new pin (`>=1.5.3`) is mostly b/c I was lazy and creating old envs to test but I'm 99% sure we can go as low as the old one (`>=0.20.3`) with the current code. I can get back to this in the afternoon and do more tests, hold on...

requirements.txt
ocefpaf: See https://github.com/ioos/erddapy/pull/300#discussion_r1158525852

requirements.txt
```diff
@@ -1,3 +1,3 @@
 httpx
-pandas>=0.20.3
+pandas>=0.25.2,<3
```
ocefpaf, line 2: @abkfenris pandas 0.25.2 is as low as we can go with the minimum Python 3.8 support we have at the moment. All the tests passes with the code change from above.

abkfenris: 4 years of compatibility is probably pretty good. I just was worried about forcing folks to 2 when the rest of their code might not be ready, or causing them to lock a version that's incompatible.

abkfenris (approved):

requirements.txt
ocefpaf: We had 1.5 when you comment. I had 2 only in the first commit b/c I thought we could not do the "trick" above and just use the non-private function instead.

ocefpaf:
The failures are due to a deprecation warning in xarray. I like the idea of making deprecation warnings errors in CIs but I couldn't figure out to limit those to erddapy only.

abkfenris:
Maybe `ignore:::xarray`?

ocefpaf:
> Maybe `ignore:::xarray`?

That works but the config is awkward b/c we need to do it for every third party library. Maybe there is a way to ensure the rule is enforced only for the main lib.

abkfenris:
`error:::erddapy.*` or similar?

PR #304, opened 2023-04-25
Changes so request_kwargs can be input to pandas

kthyng:
Currently you cannot input a keyword argument like "timeout" to urlopen because requests_kwargs and pandas_kwargs are not distinguished. Or, actually, in erddapy.py they are all just "**kw". The present changes make it so you can input keyword arguments to pandas, still not in a dict like it was, but can also specify keyword arguments to requests_kwargs. 

Notes:
* it looks like similar changes could happen in some of the other to_* functions but not sure and have not used those as much. 
* I see that requests_kwargs could instead be input to ERDDAP directly, but then that is not subsequently used. Not sure if both approaches should be allowed?


kthyng:
Example usage:

```
from erddapy import ERDDAP
server="https://erddap.sensors.ioos.us/erddap"
e = ERDDAP(server=server)
e.protocol = "tabledap"
# e.dataset_id = "cdmo_nerrs_bearcove"
e.dataset_id = "noaa_nos_co_ops_9455920"
e.to_pandas(**{"parse_dates": [0], "index_col": 0}, requests_kwargs={"timeout": 60})
```

Read in for `noaa_nos_co_ops_9455920` will only work if the input for `requests_kwargs` is working correctly.

I wasn't sure how to make a test for this that just checked to make sure the keywords made it through.

kthyng:
@ocefpaf if you get a moment could you see what you think of this?

erddapy/erddapy.py
```diff
@@ -344,10 +348,13 @@ def to_pandas(self, **kw):
...
         [1] Download a ISO-8859-1 .csv file with line 1: name (units). Times are ISO 8601 strings.
 
+        requests_kwargs: arguments to be passed to urlopen method.
+        distinct_kwargs: arguments to be passed to get_download_url and then _distinct methods
+        **kw: kwargs to be passed to third-party library (pandas).
         """
         response = kw.pop("response", "csvp")
-        url = self.get_download_url(response=response, **kw)
-        return to_pandas(url, pandas_kwargs=dict(**kw))
+        url = self.get_download_url(response=response, **(distinct_kwargs or {}))
```
ocefpaf, line 356: `distinct` should not be a `kw`. I guess we never noticed.  We should make the default `False` and make it a regular `arg`. Do you mind adding that to your PR?

kthyng: Ok I changed that in erddapy.py but didn't make changes in url.py in case that would cause some backward incompatibility issues, but I could propagate it through if you want.

kthyng:
@ocefpaf Does this look ok to you at this point? Note that I didn't change any of the other `_to_*` functions.

kthyng:
Ping @ocefpaf 😅 I'm using this locally for now but eventually will need to be able to use it otherwise so just want to keep this around.

kthyng:
@ocefpaf Might you have a chance to merge this? I hope it is ready to go.

ocefpaf:
My internal handling/passing of all the kw is still messy but I think things are a bit more consistent now.
Thanks for the PR!

PR #315, opened 2023-08-08
add request_kwargs to to_xarray

### LINKED ISSUE #314: Timeout on large requests: httpx default timeout bottleneck
When making a large request of an ERDDAP server, my responses sometimes timeout

e.g.

```python
from erddapy import ERDDAP
e_ifremer = ERDDAP(
    server="https://erddap.ifremer.fr/erddap",
    protocol="tabledap",
)
e_ifremer.dataset_id = "ArgoFloats"

e_ifremer.constraints = {
 'latitude>=': 30,
 'latitude<=':  35,
 'longitude>=':-41,
 'longitude<=': -40}

ds = e_ifremer.to_xarray()
```
Returns the error `ReadTimeout: The read operation timed out`

However, downloading the .nc direct from the server causes no issues https://erddap.ifremer.fr/erddap/tabledap/ArgoFloats.nc?&latitude%3E=30&latitude%3C=35&longitude%3E=-41&longitude%3C=-40

so its not the ERDDAP itself cancelling the request for being too large.

This appears to be caused by a default timeout of 5 seconds from httpx. I have been able to get around it by adding timeout=30 to the call

`httpx.get(url, follow_redirects=True, auth=auth, **kwargs)`

in line 20 of `erddapy/core/url.py`

While it is possible to pass `request_kwargs` including an increased timeout in the `to_pandas` method (erddapy.py line 343). This is not available in `to_xarray`. Could it be included?

callumrollo:
Following the method in #304 this adds requests_kwargs to the `to_xarray` method.

This resolves #314 

Additional tests to follow after checking it passes the current set

### DIFF
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -4,6 +4,7 @@ import functools
 from typing import Dict, List, Optional, Tuple, Union
 
 import pandas as pd
+import xarray as xr
 
 from erddapy.core.griddap import (
     _griddap_check_constraints,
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,3 +1,4 @@
 httpx
 pandas>=0.25.2,<3
 pytz
+xarray
```
ocefpaf, line 4: I'd like to keep erddapy as light as possible. Can we revert this change? When folks try to use it they do get a module not found but we can improve that with a custom error message requiring the xarray installation.

PR #319, opened 2023-10-04
fix httpx status check change

ocefpaf:
Latest `httpx` change the OK report from `None` to 200 OK status. While the current version of the tests will work with past version of `httpx` I'm pinning to modern versions for maintainability.


ocefpaf:
Docs should not take that long... Something is amiss. I'll merge this to unblock #318 and investigate it later.

PR #328, opened 2024-01-10
Fix datetime deprecation in Python >=3.11

ocefpaf:
`datetime.utcnow()` is deprecated in lieu from `datetime.datetime.now(datetime.UTC)`.

### DIFF
--- a/tests/test_erddapy.py
+++ b/tests/test_erddapy.py
@@ -1,6 +1,6 @@
 """Test ERDDAP functionality."""
 
-from datetime import datetime
+import datetime
 
 import httpx
 import pytest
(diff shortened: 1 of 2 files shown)


tests/test_erddapy.py
```diff
@@ -52,15 +52,15 @@ def test__quote_string_constraints():
         {
             "latitude": 42,
             "longitude": 42.0,
-            "max_time": datetime.utcnow(),
+            "max_time": datetime.datetime.now(datetime.UTC),
```
ocefpaf, line 55: This is new in 3.11 and we will need to wait a bit to merge this one.

PR #329, opened 2024-01-10
Support Python 3.10 and up

ocefpaf:
Python 3.10 EOL is only in 2026-10 but this will allow us to move faster, drop some code, and avoid some deprecation warnings.  I'm not in a hurry to merge this but #328 should only be merged after this one.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,12 +16,9 @@ maintainers = [
     {name = "Callum Rollo", email = "c.rollo@outlook.com"},
     {name = "Filipe Fernandes", email = "ocefpaf+erddapy@gmail.com"},
 ]
-requires-python = ">=3.8"
+requires-python = ">=3.11"
 classifiers = [
   "Programming Language :: Python :: 3 :: Only",
-  "Programming Language :: Python :: 3.8",
-  "Programming Language :: Python :: 3.9",
-  "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
   "Programming Language :: Python :: 3.12",
 ]
@@ -56,7 +53,7 @@ select = [
     "T20", # flake8-print
     "UP", # upgrade
 ]
-target-version = "py38"
+target-version = "py311"
 line-length = 79
 
 [tool.ruff.per-file-ignores]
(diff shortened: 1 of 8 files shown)


ocefpaf:
@callumrollo and @abkfenris what do you think about this one? On one hand we will be able to move faster and adopt modern practices. However, this may be too premature b/c both 3.9 and 3.10 are still supported. What do you think?

callumrollo:
I think it's reasonable to drop earlier python versions at this point. We have mature fully featured releases of erddapy that are compatible with older python versions if users still need to run <3.11

abkfenris:
I guess my concern is if an ERDDAP release breaks erddapy, are we going to backport the changes, or ask that everyone move to 3.11 to get the fixes? 

I know some of my other dependencies just added 3.11 support, and I'm pretty sure there are others that are lagging behind and keeping folks environments delayed with them.

abkfenris:
And now I'm a little horrified by seeing that [3.8 and 2.7 release cycles overlapped](https://devguide.python.org/versions/). 

How about we follow [SPEC 0](https://scientific-python.org/specs/spec-0000/) and make the minimum version 3.10 to keep pace with the rest of the scientific ecosystem? I think that allows most of the typing updates in this PR.

pyproject.toml
```diff
@@ -56,7 +53,7 @@ select = [
     "T20", # flake8-print
     "UP", # upgrade
 ]
-target-version = "py38"
+target-version = "py311"
```
abkfenris, line 56: Ruff should use the [`requires-python` under the `[project]` table](https://docs.astral.sh/ruff/settings/#target-version).

pyproject.toml
```diff
@@ -16,12 +16,9 @@ maintainers = [
     {name = "Callum Rollo", email = "c.rollo@outlook.com"},
     {name = "Filipe Fernandes", email = "ocefpaf+erddapy@gmail.com"},
 ]
-requires-python = ">=3.8"
+requires-python = ">=3.11"
```
abkfenris, line 19:
```suggestion
requires-python = ">=3.10"
```

pyproject.toml
abkfenris, line 56:
```suggestion
```

ocefpaf: Oh. Fancy, define in one place and have the tools actually look for it ;-p

Thanks!

ocefpaf:
OK. It is passing. I'll renamed the PR and check if I got all the py311 -> py310.

abkfenris (approved):

PR #338, opened 2024-03-08
Allow for empty dataset_id info URL

### LINKED ISSUE #254: Make get_info_url rerun the root info data when no dataset_id is specified.
dataset_id is a mandatory arg but we could make it optional and return the list of all datasets as the root info page.

ocefpaf:
Closes #254


ocefpaf:
The [PacIOOS](https://pae-paha.pacioos.hawaii.edu/erddap/griddap/) server is misbehaving but this PR is good to go!

PR #361, opened 2024-09-19
oddily ncCF datasets cannot be quoted

ocefpaf:
I'm starting to wonder if we do need to quote all URLs. The behavior doesn't seem to be consistent as Bob told me. Maybe we should revert all this and quote only the rare cases where `"` is present. Anyway, let's give this one last go.

### DIFF
--- a/erddapy/core/url.py
+++ b/erddapy/core/url.py
@@ -25,8 +25,12 @@ OptionalList = list[str] | tuple[str] | None
 
 def quote_url(url: str) -> str:
     """Quote URL args for modern ERDDAP servers."""
-    # We should always quote for queries.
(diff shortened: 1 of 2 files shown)


MathewBiddle:
Is this something to push upstream to ERDDAP proper?

ocefpaf:
> Is this something to push upstream to ERDDAP proper?

Not sure but in version <2.23 we could pass unquoted URLs, no problem. Now, for >=2.23, we need to quote sometimes, kind of odd but I'm not an expert on URLs standards. I do recall Bob warning me that we should always quote. However, that breaks a few corner cases, so far the search URLs and the ncCF. Hopefully that is it...

PR #381, opened 2024-12-04
Use yarl to build URLs instead of string manipulations for a more robust URL creation

ocefpaf:
The goal is to also return an `yarl.URL` object in the future but that will be a breaking change, so we'll convert them to strings for now. This should make our URLs more robust and get some nice human/quoted urls easily.


ocefpaf:
This was a nice exercise but I'm not merging this. The are some brittle parts of the URL building that require fixing before we add a new way to build them. Also, ERDDAP URLs do not fit "perfectly" in the yarl model. I'm not sure the workarounds necessary to make this work are worth it.

PR #392, opened 2025-02-04
Update pre commits

ocefpaf:
Alternative to #391

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -13,10 +13,13 @@ on:
 jobs:
   build-docs:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


ocefpaf:
The `https://opendap.co-ops.nos.noaa.gov/erddap/index.html` server is failing at the moment with:

```
Error {
    code=500;
    message="Internal Server Error: ERROR from data source: org.postgresql.util.PSQLException: ERROR: Invalid time requestesd.
  Where: PL/pgSQL function inline_code_block line 1 at RAISE";
}
```

That could be a temporary network failure and we'll merge this PR as-is. If that server doesn't come back we may need an alternative to that test though.

PR #457, opened 2026-04-28
Move back to request

ocefpaf:
I don't like the idea of reverting back to requests, in many ways httpx is better for our case. However, WASM support on requests is much easier and we do need that moving forward. While this change should be seamless to most users, I do plan on minting a new major release to avoid breaking people's workflow.


ocefpaf:
Marking this as a draft for now b/c I want to move the cassettes to a release based asset instead of baked into the repo.

ocefpaf:
It seems that `httpxyz`, `httpx`'s fork, will have support for WASM: https://codeberg.org/httpxyz/httpxyz/pulls/41

Issue #18, opened 2018-01-23
Need to quote strings in queries


rsignell-usgs:
We need to quote strings when used in ERDDAP expressions. 

This code:
```python
    kwargs = {
        'minTime<=': stop_time.value,
        'maxTime>=': start_time.value,
        'cdm_data_type=': cdm_data_type
    }

    variables = ['datasetID', 'minLongitude', 'minLatitude']
    url = e.get_download_url(dataset_id='allDatasets', variables=variables, response='csv', **kwargs)
```
produces this URL
https://erddap-uncabled.oceanobservatories.org/uncabled/erddap/tabledap/allDatasets.csv?datasetID,minLongitude,minLatitude&minTime%3C=2018-01-23T12:52:10Z&maxTime%3E=2018-01-16T12:52:10Z&cdm_data_type=TimeSeries

which gives the response:
```
Query error: For constraints of String variables, the right-hand-side value must be surrounded by double quotes. Bad constraint: cdm_data_type=TimeSeries
```

ocefpaf:
I think I have a fix for that already somewhere. Just need time to organize it in a PR here.

rsignell-usgs:
Is this a bit tricky, since time values can be strings, but should not be quoted?

rsignell-usgs:
Actually, I just did some testing, and seems that ERDDAP can handle having quotes around the time strings, so I guess we can quote all strings!

These both work:

https://erddap-uncabled.oceanobservatories.org/uncabled/erddap/tabledap/allDatasets.csv?datasetID,minLongitude,minLatitude&minTime<=2018-01-23T12:52:10Z&maxTime>=2018-01-16T12:52:10Z&cdm_data_type="Point"

https://erddap-uncabled.oceanobservatories.org/uncabled/erddap/tabledap/allDatasets.csv?datasetID,minLongitude,minLatitude&minTime<="2018-01-23T12:52:10Z"&maxTime>="2018-01-16T12:52:10Z"&cdm_data_type="Point"

https://erddap-uncabled.oceanobservatories.org/uncabled/erddap/tabledap/allDatasets.csv?datasetID,minLongitude,minLatitude&minTime%3C=%222018-01-23T12:52:10Z%22&maxTime%3E=%222018-01-16T12:52:10Z%22&cdm_data_type=Point

ocefpaf:
[...]

PR #479, opened 2026-06-10
Fix relative-constraint false positives leaving ordinary string values unquoted (#477)

### LINKED ISSUE #477: String constraint values containing "min"/"max"/"now" (e.g. station="terminal_1") are silently left unquoted, breaking ERDDAP queries
Building a tabledap download URL with a string constraint whose value merely *contains* the letters `min`, `max`, or `now` — e.g. `station="terminal_1"`, `cdm_data_type="Maximum"`, or a regex like `station=~"(WMO.*|.*minute.*)"` — silently skips the required double-quoting of the right-hand side. ERDDAP then rejects the request with HTTP 400 *"For constraints of String variables, the right-hand-side value must be surrounded by double quotes"* — the exact error the quoting feature from #18 was added to prevent.

## Environment

- erddapy `3.2.2.dev7+gebc4ebb57` (commit `ebc4ebb570effd823ca94dce9f903152a204b326`, current HEAD)
- Windows 11 Pro, Python 3.14.5
- pandas 3.0.3, numpy 2.4.6, requests 2.34.2

## Reproduction

No network needed — this is pure URL building:

```python
from erddapy import ERDDAP

e = ERDDAP(server="https://e.org/erddap", protocol="tabledap", response="csv")

cases = {
    "CARQUINEZ (no trigger)": "CARQUINEZ",
    "terminal_1 (contains 'min')": "terminal_1",
    "snowfall_stn (contains 'now')": "snowfall_stn",
    "maximum_buoy (contains 'max')": "maximum_buoy",
    "minas_basin (contains 'min')": "minas_basin",
}

for label, value in cases.items():
    url = e.get_download_url(
        dataset_id="ds", variables=["time"], constraints={"station=": value}
    )
    rhs = url.split("station=")[1]
    quoted = rhs.startswith('"') and rhs.endswith('"')
    print(f"{label:32s} -> station={rhs!r:30s} quoted={quoted}")
```

Run with `python repro.py`.

## Observed output

```
CARQUINEZ (no trigger)           -> station='"CARQUINEZ"'                  quoted=True
terminal_1 (contains 'min')      -> station='terminal_1'                   quoted=False
snowfall_stn (contains 'now')    -> station='snowfall_stn'                 quoted=False
maximum_buoy (contains 'max')    -> station='maximum_buoy'                 quoted=False
minas_basin (contains 'min')     -> station='minas_basin'                  quoted=False
```

## Expected

Every plain string value should be double-quoted (`station="terminal_1"`), as ERDDAP requires for String-variable constraints (see the ERDDAP tabledap documentation on constraints, and the original motivation in #18). Only genuine server-side relative expressions — `now`, `now-7days`, `min(longitude)+10`, `max(time)-1day`, etc. — should be passed through unquoted.

These letter sequences are common in real identifiers (station names with `min`/`max`/`now` substrings, the metadata value `Maximum`, the word `minute` inside regexes), so ordinary queries against real datasets hit this.

## Root cause

`erddapy/core/url.py:189-194` — `_check_substrings` classifies a value as a relative expression by bare substring containment:

```python
def _check_substrings(constraint: dict) -> bool:
    """Extend the OPeNDAP with extra strings."""
    substrings = ["now", "min", "max"]
    return any(
        True for substring in substrings if substring in str(constraint)
    )
```

It is consumed in two places:

- `erddapy/core/url.py:169-178` — `_quote_string_constraints` skips quoting when `_check_substrings(v)` is true, producing the unquoted RHS shown above.
- `erddapy/core/url.py:534-552` — the constraint branch of `get_download_url` also `continue`s past time parsing for any value containing those substrings.

## Proposed fix

Detect relative expressions with anchored matching instead of substring containment:

```python
import re

_RELATIVE_CONSTRAINT = re.compile(r"^now([+-].*)?$|^(min|max)\(")

def _check_substrings(constraint) -> bool:
    """Return True only for ERDDAP relative-time/variable expressions."""
    return bool(_RELATIVE_CONSTRAINT.match(str(constraint).strip()))
```

This still accepts every form the existing tests exercise (`now-50years`, `now-20years`, `min(latitude)+5`, `min(longitude)+10`, `max(depth)-23` in `tests/test_url_builder.py::test_download_url_relative_constraints` and `::test_search_url_valid_request_with_relative_time_constraints`), so no existing test expectations change. I'd add regression tests asserting that `terminal_1`, `Maximum`, and a regex containing `minute` come out double-quoted.

## Related

- Distinct from #18, which added the original quoting feature this regresses.
- Distinct from #161/#164 (and PRs #162/#213), which introduced the relative-constraint substring check — this report is about the false positives of that check, not the feature itself.
- Distinct from the `startswith("time")` family (#198, #236, #237), which concerned constraint *keys*; this is about constraint *values*.
- Open PR #474 (mypy typing) does not touch `_check_substrings` or the quoting logic, so it would not fix this.

Happy to submit a PR for this if useful.

consigcody94:
# Fix relative-constraint false positives in `_check_substrings`

## Summary

String constraint values that merely contain the letters `now`, `min`, or `max` (e.g. `station="terminal_1"`, `cdm_data_type="Maximum"`, or a regex constraint containing the word `minute`) were misclassified as ERDDAP relative (server-side) expressions and silently left unquoted. ERDDAP then rejects the query with HTTP 400: *"For constraints of String variables, the right-hand-side value must be surrounded by double quotes"* — the exact error the original quoting feature (#18) was added to prevent. This PR replaces the bare substring containment test with anchored patterns that follow the actual ERDDAP relative-constraint grammar.

Fixes #477

## Root cause

`_check_substrings` in `erddapy/core/url.py` decided a value was a relative expression via `substring in str(constraint)` over `["now", "min", "max"]`, so any ordinary value containing those letters anywhere was treated as relative and skipped both quoting and date parsing.

## What the fix changes

- `erddapy/core/url.py`: `_check_substrings` now matches the value against anchored patterns derived from the [ERDDAP tabledap constraint grammar](https://erddap.ioos.us/erddap/tabledap/documentation.html#constraints): `^now([+-].+)?$` (e.g. `now`, `now-7days`) and `^(min|max)\(.+\)([+-].+)?$` (e.g. `min(depth)`, `max(time)-1day`). Genuine relative constraints behave exactly as before; the existing relative-constraint tests (`test_download_url_relative_constraints`, `test_search_url_valid_request_with_relative_time_constraints`) are unchanged and still pass.
- `tests/test_erddapy.py`: adds `test__quote_string_constraints_substring_false_positives` (values `terminal_1`, `snowfall_stn`, `maximum_buoy`, `minas_basin`, `Maximum`, and a regex containing `minute` must all be quoted) and `test__quote_string_constraints_relative_constraints` (`now-7days`, `min(depth)`, `max(time)-1day` must not be quoted).
- `tests/test_url_builder.py`: adds `test_download_url_quotes_substring_false_positives`, an offline end-to-end check that `get_download_url` quotes such station values and regex constraints.

## Test evidence

Regression tests on the base branch (fix reverted):

```
FAILED tests/test_erddapy.py::test__quote_string_constraints_substring_false_positives
FAILED tests/test_url_builder.py::test_download_url_quotes_substring_false_positives
2 failed, 1 passed in 12.51s
```

with the failing assertion showing the malformed URL:

```
E           assert False
E            +  where False = ...endswith('&station="terminal_1"')
E            +    'https://standards.sensors.ioos.us/erddap/tabledap/org_cormp_cap2.csv?station,z&station=terminal_1'.endswith
```

With the fix applied, the offline suite (`pytest -m "not web" --ignore=tests/test_to_objects.py`, Windows 11 / Python 3.14):

```
18 passed, 24 deselected
```

(baseline on `main` before this PR: 15 passed, 24 deselected — the 3 new tests account for the difference; no existing test changed behavior). `test_to_objects.py` is excluded only because `iris` has no Python 3.14 wheel in this environment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/erddapy/core/url.py
+++ b/erddapy/core/url.py
@@ -4,6 +4,7 @@ import copy
 import datetime
 import functools
 import io
+import re
 from collections import OrderedDict
 from typing import Any, BinaryIO
 from urllib import parse
@@ -180,10 +181,22 @@ def _format_constraints_url(kwargs: dict) -> str:
 
 
 def _check_substrings(constraint: str) -> bool:
-    """Extend the OPeNDAP with extra strings."""
-    substrings = ["now", "min", "max"]
+    """Return True if the constraint is a relative (server-side) expression.
+
+    ERDDAP relative constraints are "now", optionally followed by an offset,
+    like "now-7days", or "min(varName)"/"max(varName)", optionally followed
+    by an offset, like "max(time)-1day". Those must not be quoted nor parsed
+    as dates. See
+    https://erddap.ioos.us/erddap/tabledap/documentation.html#constraints
+
+    """
+    relative_expressions = (
+        r"^now([+-].+)?$",
+        r"^(min|max)\(.+\)([+-].+)?$",
+    )
+    value = str(constraint)
     return any(
-        True for substring in substrings if substring in str(constraint)
+        re.match(expression, value) for expression in relative_expressions
     )
 
 
--- a/tests/test_erddapy.py
+++ b/tests/test_erddapy.py
@@ -67,6 +67,39 @@ def test__quote_string_constraints():
             assert value.endswith('"')
 
 
+def test__quote_string_constraints_substring_false_positives():
+    """Plain values that merely contain now/min/max must still be quoted."""
+    kw = _quote_string_constraints(
+        {
+            "station=": "terminal_1",
+            "station!=": "snowfall_stn",
+            "station<": "maximum_buoy",
+            "station>": "minas_basin",
+            "cdm_data_type": "Maximum",
+            "station=~": "(WMO.*|.*minute.*)",
+        },
+    )
+
+    for value in kw.values():
+        assert value.startswith('"')
+        assert value.endswith('"')
+
+
+def test__quote_string_constraints_relative_constraints():
+    """Relative constraints like now-7days must not be quoted."""
+    kw = _quote_string_constraints(
+        {
+            "time>": "now-7days",
+            "depth>": "min(depth)",
+            "time<": "max(time)-1day",
+        },
+    )
+
+    for value in kw.values():
+        assert not value.startswith('"')
+        assert not value.endswith('"')
+
+
 def test__format_constraints_url():
     """Test constraint formatting."""
     kw_url = _format_constraints_url(
--- a/tests/test_url_builder.py
+++ b/tests/test_url_builder.py
@@ -230,6 +230,35 @@ def test_download_url_relative_constraints(e):
     assert options["longitude<"] == max_lon
 
 
+def test_download_url_quotes_substring_false_positives(e):
+    """Test that values containing now/min/max substrings get quoted."""
+    dataset_id = "org_cormp_cap2"
+    variables = ["station", "z"]
+
+    stations = [
+        "terminal_1",
+        "snowfall_stn",
+        "maximum_buoy",
+        "minas_basin",
+    ]
+    for station in stations:
+        url = e.get_download_url(
+            dataset_id=dataset_id,
+            variables=variables,
+            response="csv",
+            constraints={"station=": station},
+        )
+        assert url.endswith(f'&station="{station}"')
+
+    url = e.get_download_url(
+        dataset_id=dataset_id,
+        variables=variables,
+        response="csv",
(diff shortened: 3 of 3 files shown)


ocefpaf:
See https://github.com/ioos/erddapy/issues/477#issuecomment-4679323953

PR #480, opened 2026-06-10
Fix get_var_by_attr never matching falsy attribute values (#478)

### LINKED ISSUE #478: get_var_by_attr never matches falsy attribute values (valid_min=0, empty strings) due to truthiness short-circuit
`ERDDAP.get_var_by_attr` can never match a variable whose attribute value is falsy — `valid_min=0`, `valid_min=0.0`, `comment=""`, `flag_values=0`, etc. The matching guard short-circuits on the attribute's truthiness before comparing it, so `e.get_var_by_attr(dataset_id, valid_min=0)` silently returns no matches even when variables carry exactly that attribute value.

## Environment

- erddapy `3.2.2.dev7+gebc4ebb57` (commit `ebc4ebb570effd823ca94dce9f903152a204b326`, current HEAD)
- Windows 11 Pro, Python 3.14.5
- pandas 3.0.3, numpy 2.4.6, netCDF4 1.7.4

## Reproduction

The matching loop is exercised in isolation here (the method itself needs a live server for `_get_variables`, but the defect is entirely in this loop). `match` below is a verbatim copy of the loop at `erddapy/erddapy.py:546-555`:

```python
def match(var: dict, kwargs: dict) -> bool:
    """Exact copy of the matching loop in ERDDAP.get_var_by_attr."""
    has_value_flag = False
    for k, v in kwargs.items():
        if callable(v):
            has_value_flag = v(var.get(k, None))
            if not has_value_flag:
                break
        elif var.get(k) and var.get(k) == v:  # <-- buggy guard
            has_value_flag = True
        else:
            has_value_flag = False
            break
    return has_value_flag


print("valid_min == 0 :", match({"valid_min": 0}, {"valid_min": 0}))      # want True
print("comment == ''  :", match({"comment": ""}, {"comment": ""}))        # want True
print("flag=='active' :", match({"flag": "active"}, {"flag": "active"}))  # want True
```

Run with `python repro.py`.

## Observed output

```
valid_min == 0 : False
comment == ''  : False
flag=='active' : True
```

## Expected

All three should be `True`. The code comments that it is "Virtually the same code as the netCDF4 counterpart" (`erddapy/erddapy.py:541`), and netCDF4's `Dataset.get_variables_by_attributes` uses a presence check plus plain equality (`hasattr(var, k) and getattr(var, k) == v`), which matches falsy values correctly. erddapy's port added a truthiness guard that diverges from those semantics.

## Root cause

`erddapy/erddapy.py:551`:

```python
elif var.get(k) and var.get(k) == v:
    has_value_flag = True
```

`var.get(k)` evaluates falsy for `0`, `0.0`, and `""`, so the equality test is never reached for those values.

## Proposed fix

```python
elif k in var and var[k] == v:
    has_value_flag = True
```

This preserves the "attribute must exist" intent (a missing attribute still does not match `None`) while allowing falsy values to match, restoring parity with netCDF4. No existing test expectations encode the buggy behavior; a small unit test covering `valid_min=0` and an empty-string attribute would lock in the fix.

Caveat on severity: this only bites when querying on attributes that legitimately hold falsy values (e.g. `valid_min=0`), which is an uncommon but valid use of the API — low severity, but the failure is silent (empty result rather than an error), which makes it easy to misread as "no such variable".

## Related

Distinct from the old closed #2 and #30, which touched `get_var_by_attr` for unrelated reasons.

Happy to submit a PR for this if useful.

consigcody94:
# Fix `get_var_by_attr` never matching falsy attribute values

## Summary

`ERDDAP.get_var_by_attr` could never match a variable whose attribute value is falsy — e.g. `valid_min=0`, `valid_min=0.0`, or an empty-string `comment` — even when the requested value compared equal. The method is documented as mirroring netCDF4-python's `get_variables_by_attributes`, which matches these values fine.

Fixes #478

## Root cause

The matching branch in `erddapy/erddapy.py` read `elif var.get(k) and var.get(k) == v:` — the leading `var.get(k)` truthiness guard short-circuits to `False` for any falsy attribute value before the equality test runs.

## What the fix changes

- `erddapy/erddapy.py`: the branch becomes `elif k in var and var[k] == v:` — a presence check plus plain equality, so falsy values match while a genuinely absent attribute still does not (including searching for `v=None`).
- `tests/test_url_builder.py`: adds `test_get_var_by_attr_falsy_attribute_values`, an offline test that stubs `_get_variables` with a fake dataset and asserts `valid_min=0` and `comment=""` match the right variable (and a truthy value still matches as before).

## Test evidence

Regression test on the base branch (fix reverted):

```
>       assert matches == ["salt"]
E       AssertionError: assert [] == ['salt']
FAILED tests/test_url_builder.py::test_get_var_by_attr_falsy_attribute_values
1 failed in 10.90s
```

With the fix applied, the offline suite (`pytest -m "not web" --ignore=tests/test_to_objects.py`, Windows 11 / Python 3.14):

```
16 passed, 24 deselected
```

(baseline on `main` before this PR: 15 passed, 24 deselected — the 1 new test accounts for the difference; no existing test changed behavior). `test_to_objects.py` is excluded only because `iris` has no Python 3.14 wheel in this environment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/erddapy/erddapy.py
+++ b/erddapy/erddapy.py
@@ -546,7 +546,7 @@ class ERDDAP:
                     has_value_flag = v(var.get(k, None))
                     if not has_value_flag:
                         break
-                elif var.get(k) and var.get(k) == v:
+                elif k in var and var[k] == v:
                     has_value_flag = True
                 else:
                     has_value_flag = False
--- a/tests/test_url_builder.py
+++ b/tests/test_url_builder.py
@@ -259,6 +259,30 @@ def test_get_var_by_attr(e):
     ]
 
 
+def test_get_var_by_attr_falsy_attribute_values(e, monkeypatch):
+    """Test that get_var_by_attr can match falsy values like 0 and ""."""
+    datasets = {
+        "fake_dataset": {
+            "salt": {"valid_min": 0, "comment": ""},
+            "temp": {"valid_min": 2.5, "comment": "in-situ"},
+        },
+    }
+
+    def _get_variables(dataset_id):
+        return datasets[dataset_id]
+
+    monkeypatch.setattr(e, "_get_variables", _get_variables)
+
+    matches = e.get_var_by_attr(dataset_id="fake_dataset", valid_min=0)
+    assert matches == ["salt"]
+
+    matches = e.get_var_by_attr(dataset_id="fake_dataset", comment="")
+    assert matches == ["salt"]
+
+    matches = e.get_var_by_attr(dataset_id="fake_dataset", valid_min=2.5)
+    assert matches == ["temp"]
+
+
 @pytest.mark.web
 @pytest.mark.vcr
 def test_download_url_distinct(e):


ocefpaf:
See https://github.com/ioos/erddapy/issues/477#issuecomment-4679323953

PR #481, opened 2026-06-11
replace the pacioos server in all examples

ocefpaf:
Let's replace everywhere and not only in the big dataset request.
Follow up to #476.


ocefpaf:
Hopefully it is a transient failure with teh ifremer server. Merging this one.

PR #487, opened 2026-06-29
refactor servers_list

ocefpaf:
In this PR we try to never load the server list unless we are about to use it and make it semi-private. Users can still initialize it if they need, but it was never intended to be a public feature.

Help mitigate https://github.com/euroargodev/argopy/issues/645.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,7 +54,7 @@ repos:
     - id: add-trailing-comma
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.15.18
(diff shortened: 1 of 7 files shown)


ocefpaf:
@jemrobinson and @gmaze would this help solve https://github.com/euroargodev/argopy/issues/645?

jemrobinson:
@ocefpaf: I haven't tested it, but yes, I think this might work. Thank you!

ocefpaf:
I'll merge and mint a new release. Note that this is kind of a breaking change, but the server list functionality was never really meant for public use anyway.

PR #508, opened 2026-09-08
Consolidate deps in pyproject.toml

ocefpaf:
This should make the project dependencies a bit easier to manage.

I was going to use pixi here originally, but it turns out it was simpler to use just PyPI packages for once.


tests/test_url_builder.py
```diff
@@ -238,10 +237,10 @@ def test_download_url_relative_constraints(e):
 
 def test_download_url_relative_constraints_non_coordinate(e):
     """Test download URL with relative constraints."""
-    dataset_id = "OBIS"
```
ocefpaf, line 241: @MathewBiddle was this dataset removed from https://erddap.ioos.us/erddap/ or am I hitting some temporary connection issues?

MathewBiddle: Yeah, I needed to pull down some of those test datasets. I didn't realize they were getting harvested upstream. 

I was back and forth on how to pull them from the harvest. We have two ways:
1. Remove the dataset from ERDDAP.  (What I chose) 
2. Explicitly don't create an ISO metadata record for harvesting. See [docs](https://erddap.github.io/docs/server-admin/datasets#iso19115file).

Should we continue hosting that test dataset for this purpose? Or, just not make ISO so the metadata doesn't get picked up?

ocefpaf: There is no need to re-add it for my tests here. I just wanted to confirm before re-writing the tests. I can rewrite them and use something else. Thanks for the clarification!

MathewBiddle: As a test, I removed the ISO metadata from this dataset: https://erddap.ioos.us/erddap/info/etopo5_EDDGridCopy/index.html

The dataset is still available, you just can't get the ISO metadata response.

PR #511, opened 2026-09-14
Loosen url check on class

### LINKED ISSUE #510: Cannot create ERDDAP with a localhot
Hi,

With CI tests at https://github.com/euroargodev/argopy we use a mocked http server to avoid having tests to fail just because some erddap server would be offline.

However, I noticed in the 3.3 release that Erddapy now raises an error with non-official Erddap server url.
Basically our mocked http server (which is a local address like `http://127.0.0.1:9898`) is not in the awesome-erddap list.

Would it be possible to implement some bypassing of servers() ?
(Happy to push a PR on this)

ocefpaf:
Fixes #510.

We still need to check if it is a valid server, if not try the Awesome ERDDAP list of acronyms, then fallback to whatever the user provided. We still need to be strict in the xarray backend though. If we aren't checking for the `/erddap` string there we may hijack OPeNDAP, https requests, etc. this is a thorn issue and we want to load intro xarray only the URLs that we are 100% sure are ERDDAP.


gmaze (commented):
Hi @ocefpaf 
An intermediate approach for very specific use-cases like https://github.com/ioos/erddapy/issues/510 would consistent in adding some kwarg to by pass the known server check.
So no change in signature and behaviour but a by-pass for third-party developers.

ocefpaf:
I thought about that, but then we have to rely on users to do that and, there are valid ERDDAP servers that break that rule. In the end, there will be an error anyway if it is not ERDDAP. That check was meant for the xarray-backend only anyway.

